PR #32, opened 2019-09-11
Fix NRE

jtorvald:
Fixed two null reference exceptions with uninstantiated arrays / collections

### DIFF
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0031.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0031.cs
@@ -1,0 +1,54 @@
+﻿using System.Collections.Generic;
+using Xunit;
+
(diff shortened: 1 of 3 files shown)


jtorvald:
@mcatanzariti agree. For me it's a bit searching on what does what and where to change. But with your input I will give it a try :-)

mcatanzariti:
You're right, this library is under documented.
It's in my plans.

However, I hope the architecture and the naming convention is clear enough to help others understand the code

jtorvald:
@mcatanzariti yeah it's probably more my experience in this area than the code itself.

jtorvald:
@mcatanzariti I wrote a test and fixed the write methods. I don't do this very often so I hope everything is fine now. Feedback is appreciated.

mcatanzariti:
Thank you for the fix!

Issue #35, opened 2019-09-12
Bug: required parameterless constructor


jtorvald:
I think this is related to #8 
I had a generic class with a constructor that takes one T parameter.
That crashed with a NRE. I wrote a test and a fix. Will do a pull request for it. Not 100% sure if that is the correct way of doing it. Feedback is welcome 👍

mcatanzariti:
The current implementation is expecting the CborConstructor attribute for non default constructors (or a manual mapping by api via ObjectMapping)

I'm not quite comfortable to change this default behavior because constructor mapping is not always straight forward.

Is it possible for you to add the CborConstructor attribute on your constructor?

jtorvald:
Like this?
[CborConstructor]
public Foo(T data)

In my case I can do that but when you use a library with you can not always do that (in that case I'd need to make a mapping?).
I look at this library as a drop-in replacement for Newtonsoft.Json and there it is possible.

I understand that this might be some tricky stuff. At least it would be good to give clear feedback to the developer what he needs to do instead of a NRE.

mcatanzariti:
fixed

PR #36, opened 2019-09-12
Make parameterless constructor not required

jtorvald:
See issue #35


mcatanzariti:
Sorry, I won't accept this pull request because it just fix the crash for writing but does not implement the detection of a non default constructor for reading.

fix implemented here: https://github.com/dahomey-technologies/Dahomey.Cbor/commit/632349b4617126e85c0880a96dc0f23888561477

jtorvald:
No worries. Your solution is better

PR #59, opened 2020-07-21
Added decimal type support to CborReader & CborWriter

rsx2048:
Have added support for 128bit decimal types to the reader & writer.
It uses type 28 in the CborPrimitive enum which is suggested by the spec to use for 128bit floating point values.
Have tried to follow your coding style as closely as possible.
Note the little->big endian conversion is slightly odd due to the way decimal.GetBits returns the internal ulong value.
Could you take a look & merge if acceptable.


mcatanzariti:

PR #70, opened 2020-11-21
Update .NET SDKs and NuGet dependencies

tibel:
Update to final .NET 5.0 SDK and update other SDKs and NuGet dependencies too.


mcatanzariti:
Thank you

PR #71, opened 2021-01-29
Implemented writing of empty arrays.

CodingBoulder:
Does this makes sense or is there an alternative way of writing empty arrays.
It was possible in version 1.4.9.


mcatanzariti:
@CodingBoulder,

Thank you very much
The test is not correct, you shoud use the Helper.TesWrite variation that expects a hex buffer and craft a hex buffer on http://cbor.me/

Thank you

CodingBoulder:
Ah yes of course otherwise the test is useless.

PR #72, opened 2021-01-30
Array and Map handling

tibel:
Moved the Array and Map handling out of CborReader and CborWriter.

This reduces the complexity of the reader and writer.
Also it reduces the coupling between converter and reader/writer.
Together that should help with maintainability of this wonderful library.

The converters will become simpler too (after a refactoring).

I hope to get feedback on this change, even if the change is not accepted.

### DIFF
--- a/src/Dahomey.Cbor/Serialization/CborReader.cs
+++ b/src/Dahomey.Cbor/Serialization/CborReader.cs
@@ -24,18 +24,6 @@ namespace Dahomey.Cbor.Serialization
         Decimal
     }
 
-    public interface ICborMapReader<TC>
-    {
-        void ReadBeginMap(int size, ref TC context);
-        void ReadMapItem(ref CborReader reader, ref TC context);
-    }
-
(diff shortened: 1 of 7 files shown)


mcatanzariti:
Hi @tibel,

I know the existing implementation may seem more complex but it allows to centralize in one place (reader & writer) the handling of the array & map SIZE reading & writing.
With your implementation, that was at a point the former implementation, the code to handle SIZE is repeated again & again.
Moreover, in my opinion, the way the CBOR specification allow to ways of implenting array & map SIZE handling should be implemented in reader & writer because it is where the details of the specification are centralized.
Converters are a more higher level of abstraction and are not aware about details of the CBOR specification.
If users want to implement a custom version of a map & an array, they just have to implement the right interface and it's pretty straightforward. The rest is hidden in the reader & writer classes.

tibel:
You are right. There is a little code duplication (3 times).
But after refactoring the converters are much simpler.

Only ObjectConverter does not really benefit.
But it was quite complex already.

PR #74, opened 2021-01-31
Streamline Map handling in CborReader

tibel:
Alternative approach to #72 (can be closed)

**Rationale**
From perspective of `CborReader` there shall be no difference between Array and Map.
So the special handling of map size should be removed.

This time only reading of map in `CborReader` is changed:
- `ReadMap()` now looks exactly the same as `ReadArray()`
- removed `remainingItemCount` as this belongs to `ObjectConverter` only

Hope this is what you had in mind.

### DIFF
--- a/src/Dahomey.Cbor/Serialization/CborReader.cs
+++ b/src/Dahomey.Cbor/Serialization/CborReader.cs
@@ -484,6 +484,17 @@ namespace Dahomey.Cbor.Serialization
             _state = CborReaderState.Start;
         }
 
+        public bool MoveNextMapItem(ref int remainingItemCount)
+        {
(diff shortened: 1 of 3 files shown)


mcatanzariti:
@tibel 

Hum no sorry.

You moved this code to ObjectConverter:

```
while (remainingItemCount > 0 || remainingItemCount < 0 && reader.GetCurrentDataItemType() != CborDataItemType.Break);
```

And that's exactly what should be hidden into CborReader.

Moreover I don't find the final result less complex. You just moved code in the wrong place in my opinion

tibel:
Ok, I understand that.

Still I think we are discussing about two topics:
1. no difference between Array and Map in CborReader
2. hiding complexity from ObjectConverter

re 1:
I hope we can agree on this one.

re 2:
Updated the pull request.

PR #89, opened 2022-06-08
Allow CborConverterAttribute on struct, interface and enum

rmja:
This PR allows CborConverterAttribute to be attached on structs, interfaces and enums. Tests are included.


mcatanzariti:
Please, could you add a test where CborConverterAttribute is used on an interface ?

rmja:
@mcatanzariti absolutely, the test is added...

rmja:
Doh:) Added...

PR #102, opened 2022-09-14
Add ReadOnlySequence<byte> buffer support to CborReader

### LINKED ISSUE #99: Feature request: Add ReadOnlySequence<byte> overload to CborReader
The current implementation of CborReader takes a `Span<byte>` of memory as the buffer. This however enforces that the data must be in one consecutive piece of memory. Adding support for `ReadOnlySequence<byte>` would allow for operation on fragment cbor memory which is really useful when the cbor data is large.

Maybe some inspiration can be found in [Utf8JsonReader](https://github.com/dotnet/runtime/blob/339a6bcc082c1ab229e40522d7bc84f9feca02ca/src/libraries/System.Text.Json/src/System/Text/Json/Reader/Utf8JsonReader.MultiSegment.cs).

rmja:
This fixes #99.

The implementation should be fairly straight forward. Let me know if there are any questions and feel free to make any suggestions, comments and/or edits.

Note when sequence is used as buffer, then ReadByteString() may allocate a buffer if the current span within the sequence does not contain the entire byte string. This is the reason why the ReadByteStringSequence() is added.

Using this overload is much more memory efficient than the current use of `ReadOnlySequenceExtensions.GetSpan()` which allocates a large array to include the entire cbor buffer if the sequence has more than one segment (which it basically almost always has).

### DIFF
--- a/src/Dahomey.Cbor.Tests/CallbackTests.cs
+++ b/src/Dahomey.Cbor.Tests/CallbackTests.cs
@@ -1,4 +1,6 @@
 ﻿using Dahomey.Cbor.Attributes;
+using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.Serialization;
 using Xunit;
 
@@ -6,7 +8,7 @@ namespace Dahomey.Cbor.Tests
 {
     public class CallbackTests
     {
-        private class ObjectWithCallbacks
+        private class ObjectWithCallbacks : IEquatable<ObjectWithCallbacks>
         {
             public int Id { get; set; }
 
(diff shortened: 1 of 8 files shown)


rmja:
Maybe someone can shed some light on why the test fails here? They all pass on my machine locally.

rmja:
@mcatanzariti Hi Michaël. Hope that this looks like something that could end up in the CborReader. Let me know if there is anything more I can do to get this merged. Thanks

rmja:
The last commit actually enables the requested inlining attributes. The runtime cannot inline a function that throws.

rmja:
One can argue that the stream based deserializer functions should use the use this overload instead, as the current call to `stream.ReadAsync(256, token)` is allocating an array large enough to contain the entire buffer in one consecutive block of memory.

PR #105, opened 2022-10-25
More write methods for string and byte string

### LINKED ISSUE #98: Consider adding CborWriter.WriteByteString(int) and WriteByteString(ReadOnlySequence<byte>)
Consider adding the following overloads to `CborWriter.WriteByteString`:

```
        public void WriteByteString(int length)
        {
            WriteInteger(CborMajorType.ByteString, (ulong)length);
        }

        public void WriteByteString(ReadOnlySequence<byte> value)
        {
            WriteInteger(CborMajorType.ByteString, (ulong)value.Length);
            foreach (var segment in value)
            {
                _bufferWriter.Write(segment.Span);
            }
        }
```

Both would be really useful for writing cbor where the entire byte string to be written is not in currently in memory / in one consecutive slice of memory. This is for example the case when the byte string to be written is from a file where the length is known in advance, but where the content is not loaded to memory.

rmja:
This fixes #98

### DIFF
--- a/src/Dahomey.Cbor.Tests/CborWriterTests.cs
+++ b/src/Dahomey.Cbor.Tests/CborWriterTests.cs
@@ -3,6 +3,8 @@ using Dahomey.Cbor.Util;
 using Xunit;
 using System;
 using System.Globalization;
+using System.Buffers;
+using System.Text;
 
 namespace Dahomey.Cbor.Tests
 {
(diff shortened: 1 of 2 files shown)


mcatanzariti:
You miss a test for the first commit (WriteByteString with sequence) !

rmja:
@mcatanzariti done.

rmja:
I went with the WriteByteStringSize instead of WriteByteStringLength because it seems as if "size" is throughout the library.

mcatanzariti:
Thanks a lot for your contributions !

PR #120, opened 2024-03-24
Add ability to set default naming convention

Odonno:
Add default naming convention property so that we can simply set a default naming convention that will be apply on every field like this:

```csharp
var options = new CborOptions
{
    DefaultNamingConvention = new SnakeCaseNamingConvention(),
};
```

It will still be overriden by the use of `CborNamingConventionAttribute`.

### DIFF
--- a/src/Dahomey.Cbor.Tests/NamingConventionTests.cs
+++ b/src/Dahomey.Cbor.Tests/NamingConventionTests.cs
@@ -1,66 +1,103 @@
 ﻿using Dahomey.Cbor.Serialization.Conventions;
 using Xunit;
-using System;
-using System.Collections.Generic;
-using System.Text;
 
-namespace Dahomey.Cbor.Tests
+namespace Dahomey.Cbor.Tests;
+
+public class NamingConventionTests
 {
-
-    public class NamingConventionTests
+    [Theory]
+    [InlineData("FooBar", "fooBar")]
+    [InlineData("fooBar", "fooBar")]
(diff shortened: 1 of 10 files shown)


mcatanzariti:
Hello,

Could you check https://github.com/dahomey-technologies/Dahomey.Cbor/commit/373266eb669fff530e5844bdabb2d18c45083219, please?
It should already implement what you want.

Thank you,

Michaël

Odonno:
Hello Michaël,

I don't understand this.

First, I don't understand the unecessary complexity of this commit. Isn't it more simple to pass a default naming policy? Secondly, it expect to pass a `Type` where you can simply pass a a `INamingConvention` which IMO is way better (in term of usability and maintainability).

Also, what is the point of `Lowercase` and `Uppercase` convention? My end goal is to have feature parity with `JsonNamingConvention` which mean having the following policies:

* default = `null` (PascalCase)
* `CamelCase`
* `LowerSnakeCase` (or pure SnakeCase)
* `UpperSnakeCase`
* `LowerKebabCase` (or pure KebabCase)
* `UpperKebabCase`

Furthermore, I will add tests to prevent regression.

Odonno:
* passing an object instead of a Type for default naming convention
* Added missing naming convention like explicited, to have feature parity with `JsonNamingPolicy`
* Updated and added some tests

I let the `Lower` and `Upper` conventions because they may be useful for some.. If you don't see the point of those, I'll let you remove the code.

PR #121, opened 2024-03-24
Read current data item

Odonno:
Added a new method called `ReadDataItem` to be able to get the byte array that contains the current data item in order to be used later, e.g. for lazy loading part of a message.

Tested on some cases, I am not sure it covers all edge cases.


mcatanzariti:
Could you please add some automated tests that use the new method, please?

Odonno:
Added some tests and update code accordingly.

PR #122, opened 2024-03-24
Fix conversion of type casting from object

Odonno:
It can happen that sometimes we don't know the exact type but it can be implicitly typed inside the library. The `ObjectConverter` currently only support map or array but not type casting from primitives for example.

This will allow us to do the following:

````csharp
var x = CborSerializer.Deserialize<object?>(...);

// later...

var y = (int)x; // or x as int for type checking
```


mcatanzariti:
Could you please add automated tests?

Thank you

Odonno:
Added theory tests. Those tests fail without the code change.

PR #124, opened 2024-04-01
Read number from string

Odonno:


mcatanzariti:
Hi,

Thank you for this contribution.
I think you could get away with CborReader.ReadSizeAndByte and the Double.Parse flavor that expects a ReadOnlySpan<Byte>. It would avoid the intermediate string allocation.

Let me know if you prefer I that I write the change.

Odonno:
Parsing from a `ReadOnlySpan<byte>` only exist in .NET 8. You do not yet support .NET 8.

There is parsing from a `ReadOnlySpan<char>` indeed, prior to .NET 5+. But nothing exist in the library to extract `ReadOnlySpan<char>`.

I will let you make the change for this I think.

mcatanzariti:
You are right.

The alternative is to use [Utf8Parser.TryParse](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.utf8parser.tryparse?view=net-7.0#system-buffers-text-utf8parser-tryparse(system-readonlyspan((system-byte))-system-double@-system-int32@-system-char)).

I'm going to take care of it

Odonno:
Small typos on exceptions (`single` instead of decimal or double)

PR #127, opened 2024-04-06
Change ReadDataItem return type to ReadOnlyMemory<byte>

### LINKED ISSUE #126: ReadDataItem return type
Hello,

It's me again.

I am working on performance improvement and more specifically memory allocation. I have usage of `ReadDataItem` on multiple occasions. I have the need for `ReadOnlySpan<byte>` (so that it stays in the stack) but I also sometimes need to pass it to another object instance. This is not possible with a `Span`. Best option is to use a `ReadOnlyMemory<byte>`. I made the change locally and it works like a charm using the inner `Span` of the `ReadOnlyMemory<byte>`.

What do you think about this change?

* How to implement multiple return types on a single method? (as it is not possible in C#)
* I may have missed some articles but there may be a better option than `ReadOnlyMemory<byte>`?

Odonno:
Fix #126

### DIFF
--- a/src/Dahomey.Cbor.Tests/CborReaderTests.cs
+++ b/src/Dahomey.Cbor.Tests/CborReaderTests.cs
@@ -327,14 +327,14 @@ namespace Dahomey.Cbor.Tests
             Assert.Equal("id", key1);
 
             var idValueItem = reader.ReadDataItem();
-            Assert.Equal(expectedIdValue, idValueItem.BytesToHex());
+            Assert.Equal(expectedIdValue, idValueItem.Span.BytesToHex());
 
             reader.MoveNextMapItem(ref remainingItemCount);
             var key2 = reader.ReadString();
             Assert.Equal("result", key2);
 
             var resultValueItem = reader.ReadDataItem();
-            Assert.Equal(expectedResultValue, resultValueItem.BytesToHex());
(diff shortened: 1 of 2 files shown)


mcatanzariti:
In my opinion ReadDataItem should return a slice of the internal CborReader buffer as a ReadOnlySpan.

Then if you want to hold a copy as a ReadOnlyMemory or an array you can always call CopyTo on the returned ReadOnlySpan buffer.

I have created a branch to reflect my proposal:
https://github.com/dahomey-technologies/Dahomey.Cbor/blob/62c4db28ab3efa3152bf4296457cb33900262893/src/Dahomey.Cbor/Serialization/CborReader.cs#L779#L790

If you are ok with this modification, I will create a PR over the new branch and I let you close this PR

Odonno:
I am ok with this. That simplify the `ReadDataItem` method at the same time.

Final question, what do thing about adding a `ReadDataItemAsMemory` that encapsulate the ReadDataItem + copy to ReadOnlyMemory logic in the library? Or maybe an extension method `ToReadOnlyMemory` or something like that? I am ok if you don't want this. Just wanted to move code out of my project as much as possible.

mcatanzariti:
If you can create an extension method, it is better I think.

PR #130, opened 2024-04-07
Turn converters in PrimitiveConverterProvider into singleton

Odonno:
Continuing work to reduce memory allocation compared to `System.Text.Json`.

I noticed huge memory usage from converters. Managed to reduced from 50+KB to ~22KB following this:

* Using `ConverterProvider` instead of registering all converters by hand, at least for those that are never used on primary execution. This is done in my own code FYI.
* Leveraging Singleton pattern for frenquently evaluated converters, via `Lazy<T>`
* Hence updating `PrimitiveConverterProvider` so that the inner converters of this provider will pass to gen 2 at some point, avoiding memory pressure

Couple of self remarks/questions:
* What about a `FrozenDictionary` for lazy converters. `string`, `bool` and `int` are accessed so frequently, I suppose it should be protifable. Not really sure for this.
* Making a self-made singleton for the `DateTimeConverter`. It cannot be put in a `Lazy<T>` and making a singleton can be another optimization. What do you think?

### DIFF
--- a/src/Dahomey.Cbor/Serialization/Converters/Providers/PrimitiveConverterProvider.cs
+++ b/src/Dahomey.Cbor/Serialization/Converters/Providers/PrimitiveConverterProvider.cs
@@ -6,31 +6,35 @@ namespace Dahomey.Cbor.Serialization.Converters.Providers
 {
     public class PrimitiveConverterProvider : CborConverterProviderBase
     {
-        private static readonly Dictionary<Type, Type> _converterTypes = new Dictionary<Type, Type>
+        private static readonly Dictionary<Type, Lazy<ICborConverter>> _lazyConverterTypes = new()
         {
-            [typeof(bool)] = typeof(BooleanConverter),
-            [typeof(sbyte)] = typeof(SByteConverter),
-            [typeof(byte)] = typeof(ByteConverter),
-            [typeof(short)] = typeof(Int16Converter),
-            [typeof(ushort)] = typeof(UInt16Converter),
-            [typeof(int)] = typeof(Int32Converter),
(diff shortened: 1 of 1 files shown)


mcatanzariti:
Thank you for this contribution.

About the DateTimeConverter why can't you put it in the dictionary of lazy objects?

The proposed implementation is not satisfying enough because the DateTimeConverter will be instantiated each time it is required.

Michaël

mcatanzariti:
I will always privilege performance over memory optimization of a few kB

mcatanzariti:
Forget what I wrote above, it is totally ok because all converters are cached in the CborConverterRegistry.
And I think because of that, we don't need to go further with FrozenDictionary

Odonno:
> Forget what I wrote above, it is totally ok because all converters are cached in the CborConverterRegistry. And I think because of that, we don't need to go further with FrozenDictionary

Ah yes, I kind of missed that too. 

`FrozenDictionary` being useless indeed. And the `Lazy<T>` is also pure overhead I suppose.

mcatanzariti:
We can improve it with a simple switch case to just instantiate what is needed. I'm going to make the change

Odonno:
> We can improve it with a simple switch case to just instantiate what is needed. I'm going to make the change

Yes, that was one of my idea at first. But then I forgot this idea. :) 
In general, if it is possible, instanciate directly instead of using `Activator`. I suppose we have to use `Activator` for everything related to generics...

mcatanzariti:
> In general, if it is possible, instanciate directly instead of using Activator. I suppose we have to use Activator for everything related to generics...
Right!

PR #136, opened 2025-03-22
Add ability to read dataitem without advancing

Odonno:
I have the need to read the current DataItem without advancing. What I want to do is:

1. Read the current data item and store it in a `ReadOnlyMemory`
2. Do not advance (or go back to the previous position)
3. Re-read the current data item but this time reading the semantic tag (and the value if needed)


mcatanzariti:

PR #143, opened 2025-09-13
Eric/only dot9

EricSites:


mcatanzariti:
Hi @EricSites,

.NET 9 definitely needs to be added 
Nevertheless, it's too soon to remove .NET 8 because it's a LTS version which end of support is scheduled for November 26.

Cheers,

Michaël

Issue #139, opened 2025-08-01
Support RFC-8742 - Concise Binary Object Representation (CBOR) Sequences - in a performant way without catching exceptions


ChristophLindemann:
It would be nice to support RFC-8742 (Concise Binary Object Representation (CBOR) Sequences) in a _sane and performant_ way ,

instead of the naive and non-performant way, which was copy pasted from [my workaround](https://github.com/dahomey-technologies/Dahomey.Cbor/issues/67#issuecomment-678929036) in #67 and dumped into the codebase (#104).

As I wrote in the original issue (#67):
> I think it can be reasonable easy be supported in CborReader by using ReadOnlySequence<byte> / SequenceReader<byte> / PipeReader.
> 
> Another benefit would be, if you have very large CBOR objects, using something like the PipeReader would enable you to -not- read the complete CBOR stream into memory before decoding.

ChristophLindemann:
Oh, and hopefully, this time I will have time to do a PR for you.

mcatanzariti:
Hi @ChristophLindemann,

To be honest, I'm not particularly fond of PipeReader architecture proposed by Microsoft a few years ago, it brings a lot of complexity in the implementation.

But it if you think it's the best solution for streaming, I'm curious to see your proposal!

Michaël

mcatanzariti:
yes, you're right the TryRead methods seems to be the right solution.

Issue #134, opened 2024-08-05
Allow streaming of CBOR objects


jamiec7919:
Hi,  not sure if this isn't already possible and I've missed it but I'd like to be able continuously perform a ReadAsync and return objects from an indefinite length stream (e.g. a network connection which is just continuously streaming objects)?

Digging in it appears that the main read functions require a fixed length stream, and the multiple reads only allow an array (and supported by my experiments which just hang!).

If it isn't possible currently do you think it is possible, maybe extending the ReadMultiple.. and is it something you'd like in the library?  Some other libraries do seem to support this, suggesting CBOR is already framed sufficiently to allow this (so I don't have to add additional framing to the stream).

mcatanzariti:
@jamiec7919,

Yes I'm curious to see your proposal!

Cheers,

Michaël

PR #154, opened 2026-08-01
CborDataItemScanner: non-throwing data item length scanning (#139, #134)

Rafael-SOWNet:
Addresses #139 (RFC 8742 CBOR sequences) and #134 (streaming). Both need the same primitive: given a buffer that may end mid-item, determine whether a complete data item is present and how many bytes it occupies — **without throwing when it isn't**.

Today the only way to answer that is to attempt a decode and catch the "Unexpected end of buffer" `CborException`. That works, but it makes truncation — an ordinary, expected condition when reading from a stream — cost an exception. That is precisely the "exception-based parsing" #139 objects to.

### API

Every entry point takes either a `ReadOnlySpan<byte>` or a `ReadOnlySequence<byte>`:

```csharp
CborDataItemScanner.Scan(buffer, out int length)                    // -> CborDataItemStatus
CborDataItemScanner.Scan(buffer, limits, out int length)
CborDataItemScanner.TryGetDataItemLength(buffer, out int length)
CborDataItemScanner.TryReadDataItem(ref buffer, out ReadOnlySpan<byte> item)
CborDataItemScanner.ScanSequence(buffer, out int consumed, out int count)

CborDataItemScanner.Scan(sequence, out long length)
CborDataItemScanner.Scan(sequence, limits, out long length)
CborDataItemScanner.TryGetDataItemLength(sequence, out long length)
CborDataItemScanner.TryReadDataItem(ref sequence, out ReadOnlySequence<byte> item)
CborDataItemScanner.ScanSequence(sequence, out long consumed, out int count)
```

The sequence overloads are the ones a streaming caller wants. `PipeReader.ReadAsync` yields a routinely multi-segment `ReadOnlySequence<byte>`, and `CborReader` is already built on sequences, so the scanner walks segment boundaries in place rather than requiring the buffer to be flattened first — which would reintroduce the copy that reading incrementally exists to avoid:

```csharp
ReadResult result = await pipeReader.ReadAsync(token);
ReadOnlySequence<byte> remaining = result.Buffer;

while (CborDataItemScanner.TryReadDataItem(ref remaining, out ReadOnlySequence<byte> item))
{
    Handle(Cbor.Deserialize<T>(item, options));
}

pipeReader.AdvanceTo(remaining.Start, remaining.End);
```

`remaining` is left at the start of the incomplete tail, so it can be handed straight back to `AdvanceTo` as the consumed position. The span overloads remain the fast path for data that is already contiguous, and a single-segment sequence is dispatched to them.

`CborDataItemStatus` separates the outcomes callers must treat differently: `Complete`, `Incomplete` (read more and retry), `Malformed` (more data will not help), `TooDeep` and `TooLarge` (resource guards).

### Bounding depth and size

`CborScanLimits` carries both bounds. `default(CborScanLimits)` is the defaults, not a pair of zeroes.

**Depth** — a buffer of `0x9F` bytes opens one indefinite-length array per byte, so an unbounded scanner is a stack-overflow vector on untrusted input. Default 64, matching `System.Formats.Cbor`.

**Item size** — a declared length is believed long before its bytes arrive. `5B FF FF FF FF FF FF FF FF` is nine bytes announcing a 16-exabyte byte string; with no bound the only available answer is `Incomplete`, which tells a streaming caller to buffer forever for bytes that will never come. `MaxItemSize` turns that into `TooLarge` on the first nine bytes, before anything is allocated. The bound is checked against declared lengths and item counts — an array costs at least one byte per item, a map two per pair — so it holds for an item whose payload has not been received. Counts are compared by division so a count near `ulong.MaxValue` cannot overflow into looking affordable. Unlimited by default, except that a length exceeding `long.MaxValue` is always refused: no `ReadOnlySequence<byte>` could hold it, so those bytes are unobtainable rather than merely absent.

### Scope

Scanning is **structural only** — it walks headers and skips payloads. It does not validate UTF-8, tag semantics or map key uniqueness, and it accepts non-preferred integer encodings. So `Complete` means "a decoder will find a whole item here", not "this item is canonical". Documented on the type.

Structural validation that *is* performed: reserved additional-information values 28–30, indefinite lengths on major types that have no indefinite form, stray break codes, indefinite-length strings whose chunks are not definite-length strings of the same major type, indefinite maps ending between a key and its value, and two-byte simple values below 32 (RFC 8949 §3.3 forbids that encoding, and a decoder rejects it, so calling it `Complete` would break the contract above).

### Tests

89 cases. Three are properties rather than examples:

* **every case is scanned three ways** — as a span, as a single-segment sequence, and as a sequence carrying one byte per segment — and all three must agree on status and length. The contiguous and segmented walks are separate implementations (a `ref struct` cannot be a generic type argument on netstandard2.0, so they cannot share one walk without pushing the contiguous case through `SequenceReader`), and this equivalence is what keeps them in step. It also means every case exercises a header, an argument and a payload split across segment boundaries.
* **every strict prefix** of a complete item reports `Incomplete` — never `Malformed`, never a bogus `Complete`. This is the property that makes the scanner safe on a stream, so it's checked exhaustively over all prefixes of 13 representative encodings rather than by sampling.
* **scanned lengths agree with what the serializer actually emits**, so a sequence walk cannot drift.

Plus the depth and size limits, absurd declared lengths that must not overflow, and sequence walking with a truncated tail in both shapes.

### Still open on top of this

The async `CborSequenceReader` over a `Stream`/`PipeReader` that #134 asks for. This is the primitive it needs, not the whole feature.

---

New types only; purely additive, no existing behaviour touched. 916 tests pass on net10.0 (827 before). All four TFMs build.

### DIFF
--- a/src/Dahomey.Cbor.Tests/CborDataItemScannerTests.cs
+++ b/src/Dahomey.Cbor.Tests/CborDataItemScannerTests.cs
@@ -1,0 +1,298 @@
+using System;
+using System.Collections.Generic;
+using Dahomey.Cbor.Serialization;
+using Dahomey.Cbor.Tests.Extensions;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests
+{
+    /// <summary>
+    /// Covers <see cref="CborDataItemScanner"/>: determining a data item's length without decoding it
+    /// and without throwing on truncated input — the primitive needed for CBOR sequences (RFC 8742,
+    /// issue #139) and for streaming (issue #134).
+    /// </summary>
+    public class CborDataItemScannerTests
+    {
+        private static CborDataItemStatus Scan(string hexBuffer, out int length)
+        {
+            return CborDataItemScanner.Scan(hexBuffer.HexToBytes(), out length);
+        }
+
+        [Theory]
+        // scalars
+        [InlineData("00", 1)]                       // 0
+        [InlineData("17", 1)]                       // 23
+        [InlineData("1818", 2)]                     // 24
+        [InlineData("190100", 3)]                   // 256
+        [InlineData("1A000F4240", 5)]               // 1000000
+        [InlineData("1B0000000100000000", 9)]       // 4294967296
+        [InlineData("20", 1)]                       // -1
+        [InlineData("3903E7", 3)]                   // -1000
+        [InlineData("F4", 1)]                       // false
+        [InlineData("F5", 1)]                       // true
+        [InlineData("F6", 1)]                       // null
+        [InlineData("F7", 1)]                       // undefined
+        [InlineData("F90001", 3)]                   // half 5.960464477539063e-8
+        [InlineData("FA47C35000", 5)]               // float 100000.0
+        [InlineData("FB3FD5555555555555", 9)]       // double 1/3
+        [InlineData("F820", 2)]                     // simple(32)
+        // strings
+        [InlineData("40", 1)]                       // h''
+        [InlineData("4401020304", 5)]               // h'01020304'
+        [InlineData("60", 1)]                       // ""
+        [InlineData("6449455446", 5)]               // "IETF"
+        // arrays and maps
+        [InlineData("80", 1)]                       // []
+        [InlineData("83010203", 4)]                 // [1, 2, 3]
+        [InlineData("8301820203820405", 8)]         // [1, [2, 3], [4, 5]]
+        [InlineData("A0", 1)]                       // {}
+        [InlineData("A201020304", 5)]               // {1: 2, 3: 4}
+        [InlineData("A26161016162820203", 9)]       // {"a": 1, "b": [2, 3]}
+        // tags
+        [InlineData("C074323031332D30332D32315432303A30343A30305A", 22)] // 0("2013-03-21T20:04:00Z")
+        [InlineData("D82763666F6F", 6)]             // 39("foo")
+        // indefinite lengths
+        [InlineData("5F42010243030405FF", 9)]       // (_ h'0102', h'030405')
+        [InlineData("7F657374726561646D696E67FF", 13)] // (_ "strea", "ming")
+        [InlineData("9FFF", 2)]                     // [_ ]
+        [InlineData("9F018202039F0203FFFF", 10)]    // [_ 1, [2, 3], [_ 2, 3]]
+        [InlineData("BF61610161629F0203FFFF", 11)]  // {_ "a": 1, "b": [_ 2, 3]}
+        public void CompleteItems(string hexBuffer, int expectedLength)
+        {
+            Assert.Equal(CborDataItemStatus.Complete, Scan(hexBuffer, out int length));
+            Assert.Equal(expectedLength, length);
+        }
+
+        /// <summary>
+        /// Every strict prefix of a complete item must report Incomplete — never Malformed and never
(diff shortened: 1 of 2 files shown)


mcatanzariti (changes requested):
Thanks for this — the scanner itself is genuinely high quality: clear code, excellent XML docs that explain the *why*, careful arithmetic (`5BFFFFFFFFFFFFFFFF` and huge array counts stay `Incomplete` without overflowing), the depth guard against the `0x9F` stack-overflow vector, and the two invariants ("every strict prefix is `Incomplete`" and "scanned lengths match serializer output") are exactly the right things to pin down. Purely additive, no existing behaviour touched. I have no concerns about the structural logic.

My request for changes is about the **input abstraction**, since that determines whether this primitive can actually carry #139/#134 rather than needing to be reworked later.

### `ReadOnlySpan<byte>` only — but the target use cases produce `ReadOnlySequence<byte>`

The scanner operates exclusively on a contiguous `ReadOnlySpan<byte>`. But #139 explicitly asks for the `ReadOnlySequence` / `SequenceReader` / `PipeReader` route, and that is the shape the streaming use cases actually hand you:

- `PipeReader.ReadAsync()` yields `ReadResult.Buffer`, a `ReadOnlySequence<byte>` that is routinely **multi-segment** and non-contiguous.
- `CborReader` in this repo is already built on `ReadOnlySequence<byte>` / `SequenceReader<byte>` (see `CborReader(ReadOnlySequence<byte>)` and `ReadSizeAndSequence`), so the streaming feature layered on top will be working in sequence-space, not span-space.

As it stands, a streaming caller with a spanning buffer has two options, both bad: flatten the whole `ReadOnlySequence` into a contiguous array before every scan (defeats the "don't hold the whole payload in memory" goal from #139 and reintroduces an allocation/copy on the hot path), or copy just the leading window with guesswork about how much is enough. Either way the primitive doesn't compose with the reader it's meant to feed.

### What I'd like to see

A scanning path that consumes a `ReadOnlySequence<byte>` (or a `ref SequenceReader<byte>`) without requiring the input to be flattened first, so it can walk across segment boundaries. The span overload is still worth keeping as the fast path for the single-segment case — I'm asking for the sequence-based API alongside it, since that's the one #139/#134 need.

A couple of smaller things worth folding in while the API is still unreleased (not blockers on their own, but cheap to address now):

- **Unbounded wait on absurd declared lengths.** `5B FF..FF` correctly returns `Incomplete`, but a streaming caller will then buffer forever for bytes that can never arrive. The doc notes this; I think the sequence-oriented API should also expose a way for the caller to cap the maximum item size so this can't become a memory-exhaustion vector on untrusted streams.
- **`Complete` vs decodability on non-preferred encodings.** Two-byte simple values in `F8 00`..`F8 1F` (values 0–31, not well-formed per RFC 8949 §3.3) scan as `Complete` but may be rejected by the decoder, so the "`Complete` ⇒ a decoder will find a whole item here" contract has a small hole. The structural-only scope is clearly documented, so at minimum I'd call this case out explicitly in the remarks; rejecting it would be even better.

Happy to keep the span implementation as-is under the hood — the ask is that the public entry point speak `ReadOnlySequence<byte>` so this lands as the primitive #139 described rather than one that needs a second pass. Let me know if you'd prefer to split the sequence API into a follow-up; if so, let's at least capture that decision in the issue so it isn't lost.

---

PR #155, opened 2026-08-01
Bound nesting depth on read and write (CborOptions.MaxDepth)

Rafael-SOWNet:
Two hazards, one mechanism — and **one deliberate behaviour change**, called out below.

### On write: a reference cycle is not representable

CBOR has no back-references, so a cycle in the object graph recurses forever:

```csharp
class Node { public Node Next { get; set; } }
var node = new Node();
node.Next = node;
Cbor.Serialize(node, buffer);   // StackOverflowException
```

### On read: a few bytes describe unbounded nesting

A buffer of `0x9F` bytes opens one indefinite-length array per byte, so hostile input can exhaust the stack.

Both were `StackOverflowException` — which cannot be caught and takes the process down. That also made them impossible to cover: such a test aborts the test host rather than failing. Both are now `CborException`.

### Implementation

Depth is tracked in `CborWriter.WriteMap`/`WriteArray` and `CborReader.ReadMap`/`ReadArray` — the high-level pairs every converter goes through — so objects, collections, dictionaries and tuples are all covered by one check. Depth is released on the way out, so wide-but-shallow documents are unaffected (there's a test for 500 siblings at depth 2 under a limit of 8).

New API:

```
CborOptions.MaxDepth                          (default 64)
CborWriter.DefaultMaxDepth                    (64)
CborWriter(IBufferWriter<byte>, int maxDepth)
CborReader(ReadOnlySpan<byte>, int maxDepth)
CborReader(ReadOnlySequence<byte>, int maxDepth)
CborWriter.MaxDepth / CborReader.MaxDepth
```

The existing constructors keep working and default to 64, matching `System.Text.Json` and `CborDataItemScanner`.

The write-side message names a reference cycle as the likely cause, since that is overwhelmingly what triggers it, and points at `MaxDepth` for data that genuinely is that deep.

### ⚠ Behaviour change

**Data nested deeper than 64 levels now throws instead of being serialized.** Raise `CborOptions.MaxDepth` to restore the old behaviour. No test in the existing suite nests that deeply, and unbounded recursion was previously a crash rather than a feature — but this is a real change and it's your call whether 64 is the right default.

### Tests

15 cases, including the reference cycle that could not previously be tested at all, a two-node cycle, the exact limit boundary, nested collections, hostile definite- **and** indefinite-length input on read, raising the limit, and that siblings do not accumulate depth.

`Helper.Write`/`Helper.Read` in the test project now pass `options.MaxDepth` through to the reader/writer they construct, matching what `Cbor.Serialize`/`Deserialize` do — otherwise the tests would silently not exercise the setting.

---

842 tests pass on net10.0 (827 before). All four TFMs build.

### DIFF
--- a/src/Dahomey.Cbor/Cbor.cs
+++ b/src/Dahomey.Cbor/Cbor.cs
@@ -304,7 +304,7 @@ namespace Dahomey.Cbor
             CborOptions? options = null)
         {
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             ICborConverter<T> converter = options.Registry.ConverterRegistry.Lookup<T>();
             return converter.Read(ref reader);
         }
@@ -315,7 +315,7 @@ namespace Dahomey.Cbor
             CborOptions? options = null)
         {
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             ICborConverter<T> converter = options.Registry.ConverterRegistry.Lookup<T>();
             return converter.Read(ref reader);
         }
@@ -327,7 +327,7 @@ namespace Dahomey.Cbor
             CborOptions? options = null)
         {
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             ICborConverter cborConverter = options.Registry.ConverterRegistry.Lookup(objectType);
             return cborConverter.Read(ref reader);
         }
@@ -339,7 +339,7 @@ namespace Dahomey.Cbor
             CborOptions? options = null)
         {
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             ICborConverter cborConverter = options.Registry.ConverterRegistry.Lookup(objectType);
             return cborConverter.Read(ref reader);
         }
@@ -351,7 +351,7 @@ namespace Dahomey.Cbor
         {
             var list = new List<T>();
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             while (reader.DataAvailable)
             {
                 ICborConverter<T> converter = options.Registry.ConverterRegistry.Lookup<T>();
@@ -367,7 +367,7 @@ namespace Dahomey.Cbor
         {
             var list = new List<T>();
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             while (reader.DataAvailable)
             {
                 ICborConverter<T> converter = options.Registry.ConverterRegistry.Lookup<T>();
@@ -384,7 +384,7 @@ namespace Dahomey.Cbor
         {
             var list = new List<object>();
             options ??= CborOptions.Default;
-            CborReader reader = new CborReader(buffer);
+            CborReader reader = new CborReader(buffer, options.MaxDepth);
             while (reader.DataAvailable)
             {
                 ICborConverter cborConverter = options.Registry.ConverterRegistry.Lookup(objectType);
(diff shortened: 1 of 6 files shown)


src/Dahomey.Cbor/Cbor.cs
```diff
@@ -503,7 +503,7 @@ public static void Serialize(
             in IBufferWriter<byte> buffer,
             CborOptions? options = null)
         {
-            CborWriter writer = new CborWriter(buffer);
+            CborWriter writer = new CborWriter(buffer, options.MaxDepth);
```
mcatanzariti, line 507: **Blocking: `NullReferenceException` on `options == null`.**

In this overload, `options ??= CborOptions.Default;` only runs inside the `else` branch (below), so `options` is still potentially `null` here when `options.MaxDepth` is dereferenced. Before this PR, `new CborWriter(buffer)` didn't touch `options`, so `Cbor.Serialize(obj, typeof(Foo), buffer)` with default options worked; now it throws an NRE — a regression on a public overload.

The generic `Serialize<T>` and every `Deserialize*` overload are fine because they run `options ??=` *before* constructing the reader/writer. Suggested fix — hoist the defaulting to the top of the method:

```csharp
public static void Serialize(object? input, Type inputType, in IBufferWriter<byte> buffer, CborOptions? options = null)
{
    options ??= CborOptions.Default;
    CborWriter writer = new CborWriter(buffer, options.MaxDepth);
    ...
```

The tests don't catch this because they all go through the generic `Helper.Write<T>`. Worth adding a case that exercises the non-generic overload without options.

---

src/Dahomey.Cbor/Cbor.cs
```diff
@@ -539,7 +539,7 @@ public static void SerializeMultiple(
...
```
mcatanzariti, line 543: Same `NullReferenceException` regression as in `Serialize(object?, Type, …)`: `options.MaxDepth` is dereferenced here, but `options ??= CborOptions.Default;` only runs inside the `else` branch below. `SerializeMultiple(input, typeof(Foo), buffer)` with default options will now throw an NRE. Hoist `options ??= CborOptions.Default;` above this line.

---

mcatanzariti (changes requested):
## Analysis of PR #155 — `CborOptions.MaxDepth`

Thanks for this — the motivation is solid and the design is clean. Both hazards are real, uncatchable `StackOverflowException`s (write-side reference cycles, read-side hostile nesting), and turning them into `CborException` is the right call. The depth counter lives in the high-level `WriteMap`/`WriteArray` and `ReadMap`/`ReadArray` pairs that every converter goes through, depth is correctly released on the way out (nice sibling-accumulation test), argument validation is present, and the default of 64 matches `System.Text.Json`. The 15 tests are well-chosen — including the cases that literally couldn't be tested before.

### 🔴 Blocking — NullReferenceException regression (2 sites)

Two **non-generic** public overloads dereference `options` before the `options ??= CborOptions.Default` (which is deferred into their `else` branch), so calling them without an explicit `options` now throws an `NRE` where it previously worked:

- `Serialize(object? input, Type inputType, IBufferWriter<byte>, CborOptions? = null)`
- `SerializeMultiple(object[] input, Type inputType, IBufferWriter<byte>, CborOptions? = null)`

Inline comments below with the one-line fix (hoist `options ??=` to the top). The generic overloads and all `Deserialize*` paths are correct. The suite misses it because every test flows through the generic `Helper.Write<T>`; a case on the non-generic overload without options would guard against regressions.

### 🟡 Minor

- The two `DeserializeMultipleAsync` hunks are indentation/line-ending-only noise with no functional change — worth reverting to keep the diff focused.
- `_depth--` runs outside a `finally`, so depth isn't released if an inner write/read throws. Harmless today since `CborWriter`/`CborReader` are single-use `ref struct`s discarded after the exception — just flagging for the future.

Once the NRE is fixed I'm happy with this. 👍

---

mcatanzariti (approved):
## Follow-up — all feedback addressed ✅

Re-reviewed at `7036514`. Every point from the previous review is resolved:

- **Blocking NRE** — `Serialize(object?, Type, …)` and `SerializeMultiple(object[], Type, …)` now hoist `options ??= CborOptions.Default;` above the `CborWriter` construction, so the optional-`options` overloads no longer dereference `null`.
- **Coverage** — four new tests exercise exactly the non-generic path that previously slipped through: default depth without options, `null` → `F6`, `SerializeMultiple` without options, and a reference cycle still bounded via the non-generic overload. That's the gap that let the bug through, now closed.
- **Diff noise** — the `DeserializeMultipleAsync` indentation-only hunks are gone.
- **`_depth--` outside `finally`** — documented as a deliberate choice (single-use `ref struct` + keeping the hot path JIT-inlineable). Reasonable.

The core design was already sound — depth tracked in the `WriteMap`/`WriteArray` and `ReadMap`/`ReadArray` pairs, released on the way out, sensible default of 64, good hostile-input tests. LGTM.

One thing that remains the maintainer's call, not a blocker: the deliberate behaviour change — data nested deeper than 64 now throws on write instead of serializing. It's opt-out via `CborOptions.MaxDepth`, well flagged in the PR description, and unbounded recursion was a crash rather than a feature, so 64 seems a fine default — just confirming it's an intentional acceptance.

---

Issue #132, opened 2024-04-14
Support native AoT


Odonno:
When trying to use this library in a native AoT context, I get the following error:

```
System.NotSupportedException: 'Dahomey.Cbor.Serialization.Converters.ObjectConverter`1[]' is missing native code or metadata. This can happen for code that is not compatible with trimming or AOT. Inspect and fix trimming and AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility
   at System.Reflection.Runtime.General.TypeUnifier.WithVerifiedTypeHandle(RuntimeConstructedGenericTypeInfo, RuntimeTypeInfo[]) + 0x75
   at Dahomey.Cbor.Serialization.Converters.Providers.ObjectConverterProvider.GetConverter(Type, CborOptions) + 0xe2
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext() + 0x74
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1, Func`2, Boolean&) + 0x7c
   at Dahomey.Cbor.Serialization.Converters.CborConverterRegistry.CreateConverter(Type type) + 0xbd
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey, Func`2) + 0xa4
   at Dahomey.Cbor.Serialization.Converters.CborConverterRegistry.Lookup[T]() + 0x25
   at Dahomey.Cbor.Cbor.Serialize[T](T, IBufferWriter`1&, CborOptions) + 0x58
   at Dahomey.Cbor.Cbor.SerializeAsync[T](T, Stream, CborOptions, CancellationToken) + 0x80
```

One solution would be to register types via `CborSerializable` with a `CborContext`, the same way it is done for `System.Text.Json`. See the example for reference:

```csharp
[JsonSourceGenerationOptions(
    AllowTrailingCommas = true,
    NumberHandling = JsonNumberHandling.AllowReadingFromString
        | JsonNumberHandling.AllowNamedFloatingPointLiterals,
    PropertyNameCaseInsensitive = true,
    ReadCommentHandling = JsonCommentHandling.Skip
)]
[JsonSerializable(typeof(Product))]
internal partial class SurrealDbWsJsonSerializerContext : JsonSerializerContext;
```

mcatanzariti:
Hi @Odonno,

[...]

PR #156, opened 2026-08-01
Native AOT support: [CborSerializable] + CborSerializerContext + source generator (#132)

Rafael-SOWNet:
Addresses #132, using the design you proposed there: a `CborSerializable` attribute plus a context, source-generated, mirroring `System.Text.Json`.

> **Note on scope/ordering:** this branch contains #153 (integer discriminators) because the generator has to cover both discriminator kinds in one pass. Happy to split, rebase, or hold this until the smaller PRs land — whatever suits review. #152, #153, #154, #155 are all independent of each other.

```csharp
[CborSerializable(typeof(Person))]
[CborSourceGenerationOptions(NamingConvention = typeof(CamelCaseNamingConvention))]
public partial class MyContext : CborSerializerContext { }

var context = CborSerializerContext.Default<MyContext>();
Cbor.Serialize(person, buffer, context.Options);
```

## Verified under a real `PublishAot=true` binary

Not analyzer output — an actual native executable:

| Path | CoreCLR | Native AOT |
|---|---|---|
| reflection (`AutoMap`) | works | **fails** |
| generated context | works | **works**, byte-identical output |

The reflection failure is exactly what #132 reports:

```
MissingMethodException: No suitable constructor found for converter type:
  'Dahomey.Cbor.Serialization.Converters.ObjectConverter`1[[Person, …]]'
  at CborConverterProviderBase.CreateConverter(CborOptions, Type)
```

`MakeGenericType` + `Activator.CreateInstance` produced an instantiation the AOT compiler never emitted code for. Trimmer annotations cannot fix that — the instantiation has to exist statically.

Measurements, including a wire-format cross-check against a source-generated CBOR library: https://gist.github.com/Rafael-SOWNet/bf4d10d325536ed1b283719693a4bfca

## How it works

The reflection path is untouched and remains the default; a context is opt-in.

Three additive pieces:

1. **`DelegateMemberMapping<T, TM>` / `DelegateStructMemberMapping<T, TM>`** — `IMemberMapping` implementations whose accessors arrive as ordinary delegates (`o => o.Name`) instead of a `MemberInfo`. `ObjectMapping<T>` was already fully programmable through public API and `IMemberMapping.MemberInfo` was already nullable, so the extension point existed.
2. **`ObjectConverter<T>(CborOptions, Func<T>? factory)`** — with a factory, the reflective `GetConstructor` + `Expression.Compile` is skipped. Required under AOT: once the ctor metadata is trimmed, `GetConstructor` returns null and deserialization fails with "Cannot find a default constructor".
3. **`Dahomey.Cbor.Generator`** — an `IIncrementalGenerator` shipped as an analyzer, emitting the above plus converter registrations.

**Throughput is on par with the reflection path** (198 vs 206 ns serialize; within noise on deserialize), and allocates ~25% less. `Expression.Compile` emits real IL that tier-1 optimises about as well as a lambda, so the reason to use a context is AOT compatibility, not speed. Full methodology in the gist — anything measured with less than ~50k warmup iterations on CoreCLR is measuring tier-0 code.

## The generator reports problems at build time

Anything it cannot handle is a **build error**, not a run-time failure — which otherwise only surfaces as `MissingMethodException` once running under AOT:

| Id | Meaning |
|---|---|
| `CBOR1001` | context is not `partial` |
| `CBOR1002` | type cannot be generated; would fall back to reflection and fail under AOT |
| `CBOR1003` | naming convention cannot be reproduced at compile time |
| `CBOR1004` | conflicting discriminator attributes |
| `CBOR1005` | member needs `[CborProperty(index)]` for this object format |
| `CBOR1006` | member has no usable setter (warning) |

Types are discovered **transitively**, so declaring `Person` also covers its nested `Address` and its `List<string>` member — only roots need declaring.

Supported: objects (nested, self-referential), structs (`ref`-taking accessors), enums, arrays, `List<T>`/`ICollection<T>`, `Dictionary<K,V>`/`IDictionary<K,V>`, `Nullable<T>`, primitives, all three object formats, both discriminator kinds, `[CborProperty]` name and index, `[CborIgnore]`, and the naming conventions.

Not yet generated, each reported as `CBOR1002` rather than silently falling back: collection *interfaces* (`IList<T>` etc., need `InterfaceCollectionConverter`), immutable collections, tuples, creator mappings, callbacks, `ShouldSerialize`, per-member custom converters. `init`-only and get-only members are emitted write-only with `CBOR1006`, because `(o, v) => o.X = v` won't compile for them — so records with `init` properties don't round-trip through a generated context yet; that needs creator-mapping support.

`CborValue` (DOM), anonymous types and `dynamic` are inherently reflective and are documented as unsupported under AOT rather than forced.

## Things worth your judgement

* **Registration order.** `ObjectConverter`'s constructor eagerly resolves a converter per member, so dependencies must be registered first; the generator emits in reverse topological order. Making that resolution lazy would remove the constraint entirely and is probably the better long-term fix, but it touches a hot path — I left it alone deliberately.
* **Naming conventions are reimplemented in the generator** (member names are baked into emitted code, so `INamingConvention` can't be called at compile time). The generated-vs-reflection byte-identity tests are what keep the two implementations in step — if they diverge, those tests fail.
* **Registration order** is no longer a live constraint on the generated path: member converters resolve on first use, so the topological emission order is belt-and-braces rather than load-bearing.
* **Packaging.** The generator ships inside the `Dahomey.Cbor` package, under `analyzers/dotnet/cs`, so a consumer needs no extra reference. It keeps `IsPackable=false` because it has no package of its own. CI packs, asserts the entry is in the `.nupkg`, and restores that package into a probe project that declares a context and checks the bytes it writes — a dll at any other path would restore fine and never load.

## No breaking changes

Everything is additive. `ObjectConverter(CborOptions)` now forwards to the new overload with a null factory, behaviour unchanged. The generator only activates for classes deriving from `CborSerializerContext`, so no existing project sees a new diagnostic.

---

1019 library tests and 35 generator tests pass on each of net8.0, net9.0 and net10.0 (827 library tests before), including a corpus that walks every type declared by every context in the test assembly and asserts the generated bytes match the reflection path and read back. All four TFMs build.

The generator is an incremental generator in the working sense as well as the interface sense: both entry points are `ForAttributeWithMetadataName`, nothing but values crosses a step boundary, and `IncrementalPipelineTests` pins that an unrelated edit produces no new output while an edit to a declared type in another file does.


mcatanzariti (changes requested):
Thanks for this — the depth of work here is obvious, and several pieces are things the library genuinely needed. The review below is long because the PR is large; most of it is actionable detail rather than objection.

**Disclosure on method:** this is a code-reading review. I was not able to build or run the test suite in my environment, so nothing below is backed by a failing test. Anything I state as a defect should be confirmed with one before you act on it.

---

## 1. Scope

The description covers the AOT generator, but the diff against `master` is three independent features:

| Feature | Files | Wire impact |
|---|---|---|
| AOT generator + `CborSerializerContext` | 8 new under `src/Dahomey.Cbor.Generator/` | additive |
| RFC 8746 typed arrays | `TypedArrayConverter`, `TypedArrayTags`, provider | **changes output** |
| RFC 8949 §4.2.1 deterministic encoding | `CborKeyComparer`, `DeterministicKeyOrder`, `CborOptions` | **changes output** |

The latter two carry their own public API surface (`TypedArrayMode`, `Deterministic`, `CborKeyComparer`) and are unrelated to AOT — the generator covering them is a consequence, not a reason to land them together. As it stands none of the three can be approved independently. Splitting would also let #157 land on its own merits.

## 2. `Deterministic` forcing `TypedArrayMode.LittleEndian`

`CborOptions.Deterministic`'s setter assigns `_typedArrayMode = LittleEndian`, and the `TypedArrayMode` setter then rejects `Never`. There is no way to get deterministic output with ordinary CBOR arrays.

RFC 8949 §4.2.1 says nothing about RFC 8746 tags. The justification in the comment — a numeric array is otherwise representable two ways — holds for any application-level tagging choice; determinism is "one data model, one encoding", and the data model for `int[]` is an array. The practical consequence is that this library's deterministic output will not match CTAP2 canonical, dCBOR, or any other deterministic implementation, which is the interop case determinism exists for. It is also not mentioned in the README's deterministic section, so a user enabling `Deterministic` silently changes the bytes of every `int[]` and `float[]`.

Suggest decoupling: leave `TypedArrayMode` alone. Any fixed `TypedArrayMode` is already deterministic.

Related gap: the guard only covers `ArrayLengthMode` and `MapLengthMode`. A `[CborLengthMode(LengthMode.IndefiniteLength)]` on a type or member still wins at `ObjectConverter.cs:451-453` and produces indefinite-length output under `Deterministic`, with no error.

## 3. The build-error guarantee does not hold

The description says unsupported shapes are reported as `CBOR1002` "rather than silently falling back". Grepping `src/Dahomey.Cbor.Generator/` returns no occurrence of `CborRequired`, `CborIgnoreIfDefault`, `DefaultValue`, `ShouldSerialize`, `CborLengthMode`, `CborConstructor`, `OnDeserializing`, `OnDeserialized`, `CborNamingConvention`, or `CborConverterAttribute`. `TypeCollector` reads only `CborIgnore`, `CborProperty`, `CborObjectFormat`, and the two discriminator attributes. Everything else is ignored silently and produces different bytes from the reflection path, not a diagnostic.

This is the promise that justifies the feature, so it seems worth closing before merge rather than after.

## 4. `RegisterConverter` is `TryAdd`, and the emitted code discards the result

```csharp
public bool RegisterConverter(Type type, ICborConverter converter)
    => _converters.TryAdd(type, converter);   // first writer wins
```

`Emitter` emits `converters.RegisterConverter(typeof(X), new ObjectConverter<X>(options));` and drops the `bool`. Note the asymmetry with `ObjectMappingRegistry.Register`, which uses `AddOrUpdate` and does overwrite — so you can end up with the generated mapping bound to a reflection converter.

Two ways in:

**(a) Mutual recursion `A ↔ B`.** `ObjectConverter<T>`'s constructor resolves members eagerly (`GenerateMemberConverter()` → `DelegateMemberMapping.ResolveConverter()` → `Lookup<TM>()`). `InDependencyOrder` breaks the cycle via `visiting` and emits `B` first; constructing `B` calls `Lookup<A>()` on an unregistered `A`, so the registry builds and caches a reflection converter for `A`, and the later generated `RegisterConverter(typeof(A), …)` returns `false` and is dropped. On CoreCLR this works, so the byte-identity tests stay green; under AOT it is the `MissingMethodException` this PR exists to remove. The self-reference break in `GetMemberNameForConverter` only covers direct `A → A`.

**(b) `new MyContext(existingOptions)`.** The protected constructor's XML doc invites supplying your own options. If those options have already serialized anything, the relevant converters are cached and every registration in `Configure` becomes a silent no-op.

Having `Configure` throw on a `false` return, or adding an overwriting registration API, would make both loud.

## 5. Four types classified as primitives have no concrete converter

`TypeCollector.IsPrimitive` returns true for `object`, `System.Guid`, `System.DateTimeOffset` and `System.Half`, so no registration and no diagnostic is emitted. None of the four has a case in `PrimitiveConverterProvider`: they fall through to `ObjectConverterProvider`, which reaches `MakeGenericType` + `Activator.CreateInstance` through `CborConverterProviderBase` — exactly the failure mode described in the second comment on this PR, and equally invisible to the AOT analyzer. `Half` is also inconsistent with its presence in the typed-array element list.

## 6. Non-public members diverge from the reflection path

`DefaultObjectMappingConvention` filters properties with `(getMethod.IsFamily || IsPrivate || IsStatic) && !IsDefined(CborPropertyAttribute)` and fields with `(fieldInfo.IsPrivate || IsStatic) && !IsDefined(CborPropertyAttribute)`. So `internal` properties and fields are serialized, `protected` **fields** are serialized, and any non-public member carrying `[CborProperty]` is serialized. `TypeCollector.EnumerateMembers` requires `Accessibility.Public` and drops all of them without a diagnostic — a silent wire-format change for anyone adopting a context.

## 7. Polymorphic subtypes are not discovered transitively

Base → derived is not a member edge, so `Dog : Animal` is never reached from `Animal`. Each subtype needs its own `[CborSerializable]`, otherwise `TryRegisterType` is never called for it, `_typesByDiscriminator` lacks the entry, and polymorphic reads fail or fall through to the fallback type from #153. No diagnostic. The description's "only roots need declaring" does not hold for hierarchies, which is the case where forgetting one is expensive.

(Discriminator registration itself is fine — `Register<T>(initializer)` reaches `Register(IObjectMapping)` which calls `DiscriminatorConventionRegistry.RegisterType`, and both conventions are registered in the registry's constructor, so `AnyConvention()` is true when the emitted `SetDiscriminator` runs.)

## 8. Types without an accessible parameterless constructor

Classified `TypeKind.Object`, registered with no factory and no creator mapping, no diagnostic — so deserialization fails at run time on a type the reflection path handled via `MapCreator`.

## 9. Emission bugs

- `spc.AddSource($"{contextSymbol.Name}.CborContext.g.cs")` keys on the simple name. Two contexts with the same name in different namespaces produce a duplicate hint name and the generator throws.
- `Emitter.Emit` writes `public partial class {Name}`: accessibility is hardcoded (an `internal` context gives CS0262) and containing types are not walked (a nested context emits a top-level type, so `Configure` stays abstract → CS0534). Generic contexts are not handled either.
- `AccessorName` derives from the simple name, so two roots with the same name in different namespaces collide (CS0102).

## 10. Incremental generator pipeline

`CreateSyntaxProvider` with a predicate matching every class that has a base list, a transform returning the `ClassDeclarationSyntax`, then `.Collect().Combine(CompilationProvider)`. Nothing cacheable enters the pipeline and combining with `Compilation` invalidates on every keystroke, so the full collect + `GetSemanticModel` + emit runs on every edit anywhere in the solution. `ForAttributeWithMetadataName` with value-equatable model records is the usual shape. `TypeModel` also holds `ITypeSymbol` in the pipeline, which RS1036/RS1038 flag — and `EnforceExtendedAnalyzerRules` is on.

## 11. Smaller items

- `Configure` never sets `options.DefaultNamingConvention`, so types resolved through the reflection fallback in the same options keep their declared names while generated types are converted — two naming conventions in one document.
- `AbstractDictionaryConverter.SortedEntries` allocates two arrays and boxes every key (`ForDictionaryKey` takes `object`) on **every write**, plus a `byte[]` per text key. Worth measuring for deterministic mode; the gist only covers the object path.
- `_deterministicMemberConvertersForWrite` and `_decorateKey` are published by plain field writes. The comment is right that the reference assignment is atomic, but without release semantics a reader on a weak memory model can observe the reference before the `List` contents. `Volatile.Write` or `Interlocked.CompareExchange` closes it.
- `NamingConventions.Apply` / `IsSupported` match on the simple type name, so a user type named e.g. `CamelCaseNamingConvention` is silently treated as the built-in one.
- Packaging is not done (`IsPackable=false`, no `analyzers/dotnet/cs`), so merging as-is ships nothing usable to package consumers.

## 12. Test coverage

The generated path is covered by three files and one non-trivial context. Not covered: polymorphic reads through a generated context, self-referential types (listed as supported), mutual recursion, non-public members, types without a parameterless constructor, `internal`/nested/generic contexts, and two same-named contexts. The 1043 passing tests are real, but they do not exercise the cases where the generated path diverges silently from reflection — which is where the risk sits.

---

## What is clearly right

- `CborKeyComparer` implements §4.2.1 correctly: major type first, then encoded-argument tier, then value; header-then-content for strings; and it reaches the full major-type-1 range down to -2^64 that `long` cannot express.
- Decorate-sort-undecorate avoids O(n log n) re-encodings, and throwing during decoration rather than inside the comparison to keep `CborException` out of `Array.Sort`'s `InvalidOperationException` wrapper is a nice touch.
- Snapshotting `MemberConvertersForWrite` once per write is the right call, and the reasoning about a half-reordered map under a fixed header is correct.
- Splitting `PeekDataItemType` / `IsBreak` out of `GetCurrentDataItemType` fixes a real latent bug — the tag-swallowing behaviour was wrong independently of typed arrays. The bookmark/rewind in the Array-format discriminator path is the same class of fix. Both would be worth landing on their own.
- `ObjectConverter(options, factory)` is minimal and backwards-compatible.

One style note for the maintainer to settle: comment density here is well above the rest of the repository, with several 10–15 line explanatory blocks. Useful in review, but worth a decision before merge.

## Suggested order

1. Split into three PRs.
2. Decouple `Deterministic` from `TypedArrayMode`, or document the coupling prominently as an interop decision.
3. Close the diagnostic gaps in §3, §5, §6, §7, §8 — that is the guarantee the feature rests on.
4. Fix §4 (silent `TryAdd`) and §9 (emission).
5. Move to `ForAttributeWithMetadataName`.
6. Resolve packaging before any release.

---

mcatanzariti (approved):
Approved. This supersedes my earlier changes-requested review.

Verified on this head, against `master` at `cea8461`:

- **The four conflicts are resolved the right way**, including the one git would not have flagged: `CborConverterRegistry` has zero occurrences of `[ThreadStatic] _underConstruction` / `IsUnderConstruction`, so the earlier #160 design is gone rather than lingering as dead code. `MemberMapping` is on `master`'s `lock (this)` version, `MemberConverter` keeps your `Func<ICborConverter<TM>>` overload, `Issue0149Cycles.cs` is `master`'s 277-line version, and `Dahomey.Cbor.csproj` carries both the signed `InternalsVisibleTo` and the analyzer packaging block.
- **The corpus actually binds for RFC 8746 now.** `TypedArrayCborContext` goes through the protected `CborOptions` constructor with `ReadWriteLittleEndian`, and the corpus copies the mode onto the reflection baseline, so an `ArrayConverter` emitted where a `TypedArrayConverter` belongs writes a plain array against the reflection path's tag 85 and fails. That closes the hole I described as loud and then had to correct as silent.
- **The `TypedArrayConverter` visibility fix is a good catch**, and on a mechanism worth naming: generated code is compiled into the consumer's assembly, so an internal type is unreachable there, and `Dahomey.Cbor.Tests` being an `InternalsVisibleTo` friend hides it from the suite entirely. I checked whether anything else in the emitter has the same problem — `ArrayConverter`, `ByteArrayConverter`, `CollectionConverter`, `DictionaryConverter`, `EnumConverter`, `NullableConverter`, `ObjectConverter`, `TypedArrayConverter`, `IMemberMapping`, `DelegateMemberMapping` and `DelegateStructMemberMapping` are all public, so that was the only one. Keeping the byte-order constructor and `ITypedArrayReader<TI>` internal is the right minimal widening.

One thing I have never been able to do, stated plainly since it bears on the weight of this approval: I have not executed the test suite at any point in this review — no .NET SDK in my environment. Every finding I raised came from reading, and green CI against the current baseline is the evidence that they pass, not my own run.

It took a lot of rounds, which is what a change this size should take. What made it converge was mostly on your side: splitting into #157, #158, #160 and #161 when asked, rebasing rather than arguing, and — the part that actually mattered — building the test machinery that makes the byte-identity claim structural. The generator harness, the incrementality assertions in both directions, the corpus scanning contexts rather than a hand list, and the pack assertion in CI are each there because a specific class of defect had already slipped through once.

#159 is stacked on this one and can move next.

---

PR #157, opened 2026-08-02
RFC 8746 typed arrays, opt-in per direction

Rafael-SOWNet:
Encodes a numeric array as one RFC 8746 semantic tag plus one byte string holding the whole array, instead of a headered item per element.

> **#161 has merged, and this is rebuilt on top of it.** The read-path fix it needs is now in `master`, so this diff is the feature and nothing else. Rebuilt by cherry-picking rather than rebasing: #161 was squash-merged, so its commit is in `master` by content but not by identity, and a rebase would have replayed it against itself.

```csharp
CborOptions options = new CborOptions { TypedArrayMode = TypedArrayMode.ReadWriteLittleEndian };
await Cbor.SerializeAsync(new[] { 1.5f, 2.5f }, stream, options);
// D8 55 48 00 00 C0 3F 00 00 20 40
```

On a `float[1000]` of sample data that is 4005 bytes against 4923, and the gap widens with the array. Element types and their little-endian tags: `sbyte` (72), `ushort` (69), `short` (77), `uint` (70), `int` (78), `ulong` (71), `long` (79), `Half` (84), `float` (85), `double` (86). `byte[]` is deliberately absent — a plain byte string is shorter and is what the format already uses — though reading tags 64 and 68 into one still works.

## Opt-in per direction

`TypedArrayMode` is a flags enum whose default is a true no-op: upgrading to a version that has typed arrays changes nothing for a caller who has not asked for them, neither the bytes written nor the documents accepted.

| Value | Effect |
|---|---|
| `Never` (default) | Neither read nor written. A tag in 64–87 is skipped like any other unrecognised tag. |
| `Read` | Typed arrays are read, in either byte order. Nothing is written as one. |
| `WriteLittleEndian` | Numeric arrays are written as little-endian typed arrays. |
| `ReadWriteLittleEndian` | Both. |

`Read` alone is the interop case: accept a peer's typed arrays without changing the bytes this side emits. Two bits are used, so a future `WriteBigEndian` can be added without changing the meaning of any existing value.

## What reads what

Writing applies to `T[]`. Reading fills any shape interchangeable with it — `List<T>`, `IList<T>`, `ICollection<T>`, `IEnumerable<T>`, `IReadOnlyList<T>`, `HashSet<T>`, `ImmutableArray<T>`, `ImmutableList<T>` — so a document written from a `float[]` member is readable by a consumer that declares it a `List<float>`. Whatever this library writes, all of them read.

The decoding lives in one place: `AbstractCollectionConverter` has no `unmanaged` constraint on its element type, so it reaches the `T[]` converter through an internal `ITypedArrayReader<TI>`, looked up only once a typed array actually arrives.

A tag whose element type does not match the target is an error rather than something to ignore — tag 86 into a `float[]` would otherwise reinterpret the bytes silently. Reserved tag 76 and the two binary128 tags are refused for the same reason.

## Two things worth knowing

`ArrayLengthMode` is ignored for an array actually written as a typed array: a byte string is definite-length, so there is no array header to make indefinite. It applies as usual to every array not written as one.

The object model has no typed array node. A `CborValue` holding one is a `ByteString` with the tag in `SemanticTag`, and `CborValueConverter` never writes `SemanticTag` back — so a DOM read-then-write drops the tag and turns the typed array into ordinary bytes. That is true of every semantic tag rather than anything this feature introduces, and is not changed here, but typed arrays make it easy to reach, so it is pinned by `TheObjectModelSeesATypedArrayAsAByteStringAndDropsItsTag` and stated in the README. It is now tracked as #162, which the README links; whoever fixes that will need to update that test.

## Also here

A separate first commit makes an indefinite-length byte or text string a `CborException` carrying the reader position, instead of a bare `NotSupportedException` that escapes the `catch (CborException)` a caller writes. RFC 8746 does not forbid a chunked byte string as the tag content, and a producer streaming a large array is where one would appear. One existing `[InlineData]` in `CborReaderTests.ReadString` changes with it; that is the only behavioural change outside typed arrays.

## Tests

`TypedArrayTests.cs`, 59 cases: every element type in both byte orders, all three object formats, struct and class members, nulls, empty arrays, polymorphism with a discriminator tag in front of a typed array member, dictionaries, tuples, collection interchangeability across nine shapes, the mode matrix, malformed and mismatched payloads, chunked payloads, nested-tag parity with `ArrayConverter`, and the size comparison.

The byte-swapping paths take the host byte order as a constructor argument rather than reading `BitConverter.IsLittleEndian`, so both orders are exercised on little-endian CI instead of half the code never running. That constructor is not public API, so the converter is `internal` and the test project is a signed `InternalsVisibleTo` friend, as #158 does for `CborKeyComparer`.

**1163 tests green on .NET 8, 9 and 10**, on this head, upstream — `master`'s 1104 plus 59 here:
https://github.com/dahomey-technologies/Dahomey.Cbor/actions/runs/31093022309

## One resolution worth calling out

`CborOptions` was the only real conflict, because #158 landed `Deterministic` there while this was out of tree. The two are kept **independent**: `Deterministic` does not force `TypedArrayMode`, and `TypedArrayMode` does not constrain `Deterministic`.

That is deliberate rather than the lazy resolution, and it is the §2 objection from #156 not being reintroduced. A typed array and a plain CBOR array are two encodings of the same value, but which one gets written is fixed by the setting rather than chosen per value — so either mode yields exactly one encoding and §4.2.1 is satisfied without either option constraining the other. The reasoning is on the property so the next person does not have to re-derive it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ High-performance [CBOR](https://cbor.io/) serialization framework for .Net (C#)
 * Support for collection interfaces: IList<>, ICollection<>, IEnumerable<>, IReadOnlyList<>, IReadOnlyCollection<>
 * Support for dynamics
 * Support for structs
+* Support for RFC 8746 typed arrays, opt-in per direction (CborOptions.TypedArrayMode)
 
 ## Installation
 ### NuGet
@@ -90,6 +91,63 @@ CborObject obj = new CborObject
 await Cbor.SerializeAsync(cborObject, stream);
 
 ```
+
+### Typed arrays (RFC 8746)
+
+A numeric array can be encoded as an [RFC 8746](https://www.rfc-editor.org/rfc/rfc8746) typed array: one
+semantic tag and one byte string holding the whole array, instead of a headered item per element.
+
+```csharp
+CborOptions options = new CborOptions { TypedArrayMode = TypedArrayMode.ReadWriteLittleEndian };
+await Cbor.SerializeAsync(new[] { 1.5f, 2.5f }, stream, options);
+// writes D8 55 48 00 00 C0 3F 00 00 20 40
+```
+
+Supported element types, with their little-endian tag: `sbyte` (72), `ushort` (69), `short` (77),
+`uint` (70), `int` (78), `ulong` (71), `long` (79), `Half` (84), `float` (85) and `double` (86).
+`byte[]` is deliberately not included: a plain CBOR byte string is shorter and is what the format
+already uses; reading tags 64 and 68 into a `byte[]` still works.
+
+On a `float[1000]` of realistic sample data this writes 4005 bytes against 4923 for the plain form, a
+saving of 918. The 5 bytes of overhead are `D8 55` (tag 85) plus `59 0F A0` (byte string, length 4000).
+
+#### Reading and writing are separate
+
+`TypedArrayMode` is a flags enum, and the default `Never` is a true no-op — upgrading to a version that
+has typed arrays changes nothing for a caller who does not ask for them.
+
+| Value | Effect |
+|---|---|
+| `Never` (default) | Neither read nor written. A tag in 64–87 is skipped like any other unrecognised tag. |
+| `Read` | Typed arrays are read, in either byte order. Nothing is written as one. |
+| `WriteLittleEndian` | Numeric arrays are written as little-endian typed arrays. |
+| `ReadWriteLittleEndian` | Both. |
+
+`Read` on its own is the interop case: accept a peer's typed arrays without changing the bytes this side
+produces.
+
+#### What reads what
+
+Writing typed arrays applies to `T[]`. Reading one fills any of the shapes that are interchangeable with
+`T[]` — `List<T>`, `IList<T>`, `ICollection<T>`, `IEnumerable<T>`, `IReadOnlyList<T>`, `HashSet<T>`,
+`ImmutableArray<T>` and the rest — so a document written from a `float[]` member is readable by a
+consumer that declares that member as a `List<float>`.
+
+A tag whose element type does not match the target is an error rather than something to ignore: reading
+tag 86 (binary64) into a `float[]` throws, which catches a mismatch that would otherwise silently
+reinterpret the bytes.
+
+#### Two things to know
+
+`ArrayLengthMode` has no effect on an array actually written as a typed array — a typed array is a single
+definite-length byte string, so there is no array header to make indefinite. It applies as usual to every
+array not written as one.
+
+The object model does not model typed arrays. A `CborValue` holding one is a `ByteString` with the tag in
+`CborValue.SemanticTag`, so DOM code sees opaque bytes rather than numbers. Note that
+`CborValueConverter` does not write `SemanticTag` back — reading a document into a `CborValue` and
+serializing it again drops the tag, turning the typed array into a plain byte string. That is true of
+every semantic tag, not just these, but typed arrays are the easiest way to meet it.
 
 ### Custom converters
 
--- a/README.md
+++ b/README.md
@@ -147,7 +147,8 @@ The object model does not model typed arrays. A `CborValue` holding one is a `By
 `CborValue.SemanticTag`, so DOM code sees opaque bytes rather than numbers. Note that
 `CborValueConverter` does not write `SemanticTag` back — reading a document into a `CborValue` and
 serializing it again drops the tag, turning the typed array into a plain byte string. That is true of
-every semantic tag, not just these, but typed arrays are the easiest way to meet it.
+every semantic tag, not just these, but typed arrays are the easiest way to meet it. Tracked as
+https://github.com/dahomey-technologies/Dahomey.Cbor/issues/162.
 
 ### Custom converters
 
--- a/src/Dahomey.Cbor.Tests/CborReaderTests.cs
+++ b/src/Dahomey.Cbor.Tests/CborReaderTests.cs
@@ -260,7 +260,7 @@ namespace Dahomey.Cbor.Tests
 
         [Theory]
         [InlineData("63666F6F", "foo", null)]
-        [InlineData("7F6166616F616FFF", "foo", typeof(NotSupportedException))]
+        [InlineData("7F6166616F616FFF", "foo", typeof(CborException))]
         [InlineData("F6", null, null)]
         [InlineData("C6F6", null, null)] // semantig tag 6 + null
         [InlineData("60", "", null)]
--- a/src/Dahomey.Cbor.Tests/TypedArrayTests.cs
+++ b/src/Dahomey.Cbor.Tests/TypedArrayTests.cs
@@ -1,0 +1,861 @@
+using Dahomey.Cbor.Attributes;
+using Dahomey.Cbor.ObjectModel;
+using Dahomey.Cbor.Serialization;
+using Dahomey.Cbor.Serialization.Converters;
+using Dahomey.Cbor.Tests.Extensions;
+using Dahomey.Cbor.Util;
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests
+{
+    public class TypedArrayTests
+    {
+        private static CborOptions TypedArrayOptions()
+        {
+            return new CborOptions { TypedArrayMode = TypedArrayMode.ReadWriteLittleEndian };
+        }
+
+        [Fact]
+        public void WriteFloatArrayAsTypedArray()
+        {
+            // D855 tag(85) 48 bytes(8) -> 1.5f, 2.5f little endian
+            Helper.TestWrite(new[] { 1.5f, 2.5f }, "D855480000C03F00002040", null, TypedArrayOptions());
+        }
+
+        [Fact]
+        public void WriteDoubleArrayAsTypedArray()
+        {
+            // D856 tag(86) 48 bytes(8) -> 1.5d little endian
+            Helper.TestWrite(new[] { 1.5d }, "D85648000000000000F83F", null, TypedArrayOptions());
+        }
+
+        [Fact]
+        public void WriteInt16ArrayAsTypedArray()
+        {
+            // D84D tag(77) 44 bytes(4) -> 1, -2 little endian
+            Helper.TestWrite(new short[] { 1, -2 }, "D84D440100FEFF", null, TypedArrayOptions());
+        }
+
+        [Fact]
+        public void WriteEmptyArrayAsTypedArray()
+        {
+            // D855 tag(85) 40 bytes(0)
+            Helper.TestWrite(new float[0], "D85540", null, TypedArrayOptions());
+        }
+
+        [Fact]
+        public void WriteNullArrayIsStillNull()
+        {
+            // F6 null
+            Helper.TestWrite<float[]>(null, "F6", null, TypedArrayOptions());
+        }
+
+        [Fact]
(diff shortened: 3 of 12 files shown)


mcatanzariti (changes requested):
Thanks for this — the diagnosis behind the read-path section is correct, the RFC 8746 tag table is right, and the test suite is more thorough than most contributions this size. The problems below are real, though, and a few of them are not visible from the PR description.

Findings are ordered by severity. Two caveats on method: this is a static review (I read the code against `master`, I did not execute the suite), and B1/B3 in particular are established by reading rather than by a failing test — I'd suggest confirming them locally before acting.

---

## 🔴 Blocking

### B1 — `List<float>` cannot read back a typed array this library itself wrote

`TypedArrayConverterProvider` only recognises `type.IsArray`. Every other collection goes through `AbstractCollectionConverter.Read` → `ReadNull()` (whose `SkipSemanticTag` eats tag 85) → `ReadArray` → `ReadBeginArray` → `Expect(CborMajorType.Array)` against a **byte string** → `CborException`.

Concretely: `class A { float[] Samples }` serialized with `TypedArrayMode.LittleEndian`, read back by a consumer declaring `class A { List<float> Samples }` → fails. That works today, because `T[]`, `List<T>`, `IList<T>`, `HashSet<T>` and `ImmutableArray<T>` are interchangeable when reading an ordinary CBOR array. The feature breaks that symmetry.

**The tests don't cover this, and read as if they do**: `TypedArrayInAListRoundTrips` exercises `List<float[]>` — a list *of* typed arrays. Same for `TypedArrayAsDictionaryValueRoundTrips` (`Dictionary<string, float[]>`). Nothing reads a typed array into anything other than a `T[]`.

Two ways out: lift tag 64–87 handling into `AbstractCollectionConverter`, or scope the feature explicitly to `T[]` — but then writing must be restricted too, otherwise the library emits documents it cannot read back into an equivalent shape.

### B2 — Reading is hard-enabled, so the behaviour change is imposed on every existing user

`TypedArrayMode` governs writing only ("Reading […] is always supported and is not affected by this setting", `CborOptions.cs`). On a package upgrade alone, with no caller change:

- ten array types switch converter (`ArrayConverter<T>` → `TypedArrayConverter<T>`);
- new `CborException`s become possible on documents that were accepted before — `WrongElementTypeThrows`, `ReservedTag76Throws` and `Binary128TagThrows` are exactly those cases;
- the cost in B4 applies.

`TypedArrayMode.Never` should mean *nothing changes*. Either `Never` disables reading too, or reading needs its own separate switch.

### B3 — Nested semantic tags regress; undocumented and untested

`SkipSemanticTag()` (`CborReader.cs:960`) skips exactly **one** tag. `GetCurrentDataItemType()` skipped **all** of them, via the `case SemanticTag: Advance(1); return GetCurrentDataItemType();` recursion (`:196-198`). Before this PR the probe consumed the whole stack; now `PeekDataItemType`/`IsBreak` restore, so only the converter's single `SkipSemanticTag` runs.

Case that worked and now throws: a member encoded `C1 C0 <value>` — two short nested tags. The second tag stays in the stream and `Expect` fails.

`ReadArrayFormatObjectWhoseFirstItemIsATaggedTuple` carries the comment "Both tags at once, nested", but the payload is `D827 654E616D6564 C1 820102`: those are **two sequential tags on two different items** (tag 39 plus its discriminator value, then tag 1 on the tuple), not two nested tags on one item. The comment is inaccurate and the case is uncovered.

Worth knowing while you're in there: for a nested tag with a multi-byte encoding (`C1 D8 55 …`), `master` was **already** broken — `Advance(1)` skips the header byte but not the argument. Probably better to fix nesting properly in `SkipSemanticTag` than to leave two different broken behaviours.

### B4 — Unmeasured cost on the hot path of *every* object deserialization

`PeekDataItemType()` does `GetBookmark()` + `ReturnToBookmark()` **per member read**. `CborReaderBookmark` (`CborReader.cs:61`) holds a `ReadOnlySpan<byte>`, a `ReadOnlySequence<byte>?` and a `SequenceReader<byte>` plus four fields — on the order of 120–150 bytes, copied twice, for every member of every object, whether or not typed arrays are in use. "Not benchmarked" isn't enough for this one.

There's a bookmark-free alternative in the spirit of the `IsBreak()` you already added: an `IsNull()` that inspects the current header without skipping a tag, O(1). The only functional difference is a `tag + null` member, where the probe would no longer see the null — but the member's own converter calls `ReadNull()`, which skips the tag and returns null anyway; only `RequirementPolicy.DisallowNull` enforcement on that exotic case would change.

### B5 — Bare `NotSupportedException` on a chunked byte-string payload

RFC 8746 does not forbid an indefinite-length byte string as the tag content. `ReadByteString()` → `ReadSizeAndBytes` (`:851-856`) → `size == -1` → `ThrowNotSupported()` → `throw new NotSupportedException()` with **no message** (`:1175`). It isn't a `CborException`, so it escapes the `catch (CborException)` every consumer writes.

Pre-existing limitation, but this feature exposes it precisely on its own use case — large arrays, therefore streaming, therefore chunking. Untested.

---

## 🟠 Major

**M1 — Scope.** A global read-path refactor and a feature in one PR. `CborReader` (4 sites), `MemberConverter` (2), `ObjectConverter` (1) and `TupleConverter` (7 arities) affect every user; `TypedArray*` affects nobody until it's switched on. Taking you up on your own offer: split it. Read-path fix first, with its own regression tests and B3/B4 addressed, feature on top.

**M2 — The DOM sees a tagged byte string, not an array.** `CborValueConverter.Read:49` captures the tag into `SemanticTag`, then reads a `ByteString`. The round-trip is lossless, but any code inspecting or comparing `CborValue` sees a typed array as opaque. Combined with B2, a peer that starts emitting typed arrays silently changes behaviour for DOM consumers. Worth a decision and a line in the README.

**M3 — The big-endian paths never execute.** `ReverseElements` on write and the `bigEndian != !BitConverter.IsLittleEndian` branch on read are dead code on any CI machine. They become testable if byte order is injected instead of read straight from `BitConverter.IsLittleEndian`.

**M4 — `TypedArrayMode` is global, with no granularity.** Once enabled, *every* `int[]`/`double[]`/… in the graph switches. No per-member or per-type opt-in. This is public API that can't be refined later without a break.

**M5 — `ArrayLengthMode = IndefiniteLength` is silently ignored** while typed mode is active. Legitimate — a byte string is definite-length — but silent and undocumented.

**M6 — `TypedArrayConverter<TI>` is `public` but unusable from outside.** Its table (`TypedArrayTags`) is `internal` and its constructor throws for any `TI` outside that table. Make it `internal`, or expose the table.

---

## 🟡 Minor

- `_options` is redeclared in `TypedArrayConverter` although `ArrayConverter` already has one (`private`) — promote that to `protected` rather than duplicating.
- `DescribeTag`: 24 `case` arms serving one error message.
- `TypedArrayTags.MinTag`/`MaxTag` are `public` inside an `internal` class.
- `TypedArrayConverter.Read` swallows any tag outside 64–87 and then `base.Read` skips another one, making it *more* lenient about nested tags than `ArrayConverter` — the inverse inconsistency of B3.

## ⚪ Cosmetic

- 17 commits to squash.
- Comment-to-code ratio is well above the surrounding style: `TypedArrayConverter.Read` and `IsBreak` carry six lines of commentary for two lines of code.
- The `ReadArrayFormatObjectWhoseFirstItemIsATaggedTuple` comment is factually wrong (see B3).

---

## Verified as correct

Recording what I checked, so it doesn't get re-litigated:

- **The RFC 8746 table** — all ten types × both byte orders checked individually against §2. No errors, including `sbyte` where LE == BE == 72.
- **Provider placement is right.** `_providers` is a `ConcurrentStack`, so registering after `CollectionConverterProvider` correctly puts it ahead of `ArrayConverter`, while leaving `PrimitiveConverterProvider` (`byte[]` → `ByteArrayConverter`, `:54`) and `ByAttributeConverterProvider` with priority.
- **The netstandard2.0 `Half` polyfill** is `[StructLayout(Sequential)]` over a single `ushort` → 2 bytes, `unmanaged`. `MemoryMarshal`-safe across all four TFMs.
- The "`ReadByteString` always returns a contiguous span" comment is **accurate**: `GetBytes(..., allowScratchBuffer: false)` copies into a fresh array when `UnreadSpan` is too short (`:1141-1146`).
- `Clone()` before `ReverseElements` on write — the caller's array is not mutated.
- No memory amplification: the allocation is sized from the byte string already read, not from a declared count.
- Strict rejection of non-matching tags in 64–87 is a genuine diagnostic improvement over current behaviour.

---

The work is serious and the underlying analysis is sound. What blocks merge is B1 (a real functional defect), B2 (a change imposed without opt-in), and M1 (scope, which is what makes the rest hard to review properly). Splitting the read-path fix into its own PR — as you offered — is the right next step; I'd rather review and merge that on its own merits first.

---

mcatanzariti (approved):
Approving. All five blockers are resolved, and I checked the rebuild rather than assuming the cherry-pick carried everything across.

## Rebuild verified

`TypedArrayConverter.cs`, `TypedArrayTags.cs`, `AbstractCollectionConverter.cs` and `TypedArrayTests.cs` are byte-identical to the tree I reviewed before #161 merged. Only `CborOptions.cs`, the csproj and the README differ, which is exactly the conflict surface and the #162 link. Nothing was lost or quietly reworked in the move.

Cherry-picking rather than rebasing was the right call, and for the reason you give: #161 was squash-merged, so its commit is in `master` by content but not by identity. A plain `git rebase` replays it against itself and conflicts on every hunk.

## The blockers

| | |
|---|---|
| B1 — collections | `ITypedArrayReader` seam, lazy lookup, nine shapes covered. `float[]` written / `List<float>` read is pinned. |
| B2 — opt-in | Flags enum per direction. `Never = 0` is a genuine no-op, and `Read` alone covers the interop case I had wanted default-on for. Better than what I asked for. |
| B3 | Withdrawn; pinned as a pre-existing limitation. |
| B4 — perf | `IsNull()`, no bookmark, landing under `master`. Now in via #161. |
| B5 — chunked | `CborException` with reader position and an actionable message, on all three sites. |

M2 through M6 and the minors are all addressed. The `_options` promotion to `protected` is there too.

## On the `Deterministic` resolution

This is the one place the merge order forced a judgement call that neither PR could have anticipated alone, so I checked it rather than waving it through — and the independence you chose is right.

The instinct is to reject `WriteLittleEndian` under `Deterministic`, by analogy with `RejectIndefinite`. That would have been wrong twice over. `RejectIndefinite` exists because §4.2.1 *requires* definite lengths — spec conformance, not encoding stability. And `Deterministic`'s own rule ("refuses any setting that admits more than one encoding of the same value") does not catch `TypedArrayMode`, because the choice is fixed by the setting rather than made per value, which is precisely the argument you put on the property.

The residual worry — two parties both deterministic but differing on `TypedArrayMode` produce different bytes — is real but applies identically to `ObjectFormat`, `EnumFormat`, `DateTimeFormat` and `DiscriminatorPolicy`, none of which are constrained. `Deterministic` pins the encoding form for a given value and configuration; it has never pinned configuration agreement between parties. Constraining `TypedArrayMode` alone would have been an inconsistent special case.

Putting the reasoning on the property rather than in this thread is the right instinct — it is where the next person will look.

## Not blocking

Two things stay open by design, both correctly scoped out:

- #162, the object-model tag loss. Broader than typed arrays, and `TheObjectModelSeesATypedArrayAsAByteStringAndDropsItsTag` will need updating when it is fixed — which the description already says.
- Nested semantic tags, which you offered to fix against `master` separately. Still worth doing: the double-advance in `GetCurrentDataItemType`'s `SemanticTag` case is a real bug, just not this PR's.

Thanks for the way this went — the pushback on B2 and B3 made the result better than if you had implemented all five as filed.

---

PR #158, opened 2026-08-02
Deterministic encoding: CborOptions.Deterministic (RFC 8949 §4.2.1), closes #148

### LINKED ISSUE #148: Support for RFC 8949 Section 4.2 / Deterministic Encoding / Canonical Encoding
Are there any plans to support deterministic encoding as described in RFC 8949 section 4.2?

Rafael-SOWNet:
Closes #148.

Adds `CborOptions.Deterministic`, producing RFC 8949 §4.2.1 canonical output so a document can be hashed for integrity or deduplication.

## What was already true

Three of the four §4.2.1 core requirements are already satisfied by `CborWriter` and needed no change:

- shortest-form arguments for integers, lengths and tags (`WriteInteger`'s 24/25/26/27 ladder)
- preferred float serialization — binary16, then binary32, then binary64, shortest that round-trips exactly
- definite lengths, which are already the default

So this PR adds the fourth — bytewise map-key ordering — plus refusal of any setting that admits more than one encoding of the same value.

## What it does

- Map keys sort bytewise on their **encoded** form. A shorter encoded key therefore always sorts before a longer one (`"z"` before `"aa"`), and negative integer keys sort after every non-negative one because they are major type 1. This is §4.2.1 ordering, not the deprecated §4.2.3 length-first variant.
- Ordering applies wherever a map is written: `StringKeyMap` and `IntKeyMap` objects, `Dictionary<K,V>`, and `CborObject`. `CborObjectFormat.Array` is positional and has no keys, so it is deliberately left untouched.
- An indefinite-length map or array throws a `CborException` at the point its header would be written, whichever of the three sources asked for it: `ArrayLengthMode`, `MapLengthMode`, or a `CborLengthMode` attribute on a type or member, which outranks both options. The refusal is at the writer rather than at the property setters so no configuration order can slip past it.
- Any key kind whose converter can write it is orderable, because each key is encoded through that converter and the encoded bytes are what get compared. That covers text, byte strings, every integral type, `char`, enums, floating point, booleans and `DateTime`, and for `CborObject` also nulls, arrays and nested maps, mixed freely within one map. Mixed kinds order by major type first, matching the bytewise order of the leading byte. A custom key converter participates with nothing added here.

## Cost

Ordering is decided once per write, not per member. Steady-state allocation is unchanged on both paths; a 32-member object write measured 555-661 ns non-deterministic and 582-600 ns deterministic, both at or below the pre-change baseline. Dictionary key bytes are encoded once per key rather than once per comparison — decorate-sort-undecorate — which took a 2000-entry map from 1,048,488 to 144,208 bytes.

Output with `Deterministic = false` is byte-identical to `master`, verified by `DeterministicByteCompatibilityTests`: a seventeen-shape corpus covering all three object formats, nesting, a polymorphic discriminator, dictionaries keyed by string, int, long, char, aliased and flags enums, double, bool and `byte[]`, and mixed-kind `CborObject`s. The expected bytes were captured by running that same corpus on the base commit, not written by hand.

## Public API

`CborOptions.Deterministic`, and nothing else. `CborKeyComparer` is `internal`: comparing encoded bytes left it with two live methods serving `ObjectConverter`'s member sort, and its old cross-kind signature was an artifact of the previous approach rather than something worth freezing.

The test project is signed with the library's own key so it can be an `InternalsVisibleTo` friend — a strong-named assembly can only grant friend access to another strong-named one. Without it the choice would have been between keeping the class public and dropping the direct tests for a 65536-character key and for `int.MinValue`, neither of which the encoded bytes can reach.

## Known limitations, stated rather than implied

- Enum keys are ordered by the bytes `EnumConverter` actually emits, including its existing 32-bit reinterpretation of wider underlying types. An order derived from a value the document does not contain would be a lie about the document — but it means two enum values wider than 32 bits can collide into duplicate keys. The README says so.
- The flag must not be changed while a write using those options is in flight.

## Tests

974 on `origin/master` → 1080 here, 0 failures; all four TFMs build, and CI is green on all three runnable ones. Coverage includes the encoded-size tier boundaries (23/24, 255/256, 65535/65536) for both text and integer keys, `int.MinValue` and `ulong.MaxValue`, negative `IntKeyMap` indices, discriminated hierarchies, mixed-kind `CborObject` maps, a regression test for changing the flag mid-write, nested maps sorted independently of their container, and non-scalar `CborObject` keys ordered rather than rejected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,22 +227,45 @@ sorts before ``"aa"``. Keys of different CBOR major types order by major type fi
 integer, then negative integer, then byte string, then text string — which is why negative integer
 keys sort after all non-negative ones.
 
-Supported dictionary key types are ``string``, ``char``, ``byte[]``/``ReadOnlyMemory<byte>``, every
-integral type (``byte``, ``sbyte``, ``short``, ``ushort``, ``int``, ``uint``, ``long``, ``ulong``)
-and enums — each ordered as its own converter writes it. For an enum that means: with
-``EnumFormat.WriteToString`` a value that *has* a name is written and ordered as that name, and a
-value that has none (a combination of flags, or a cast from an unlisted number) falls back to the
-integer form in both. ``CborObject`` keys may be text strings, byte strings and integers, mixed
-freely within one map, which is what lets a document read off the wire be re-encoded
-deterministically. Any other key type throws a ``CborException`` rather than silently emitting
-unsorted output.
+Any key type whose converter can write it can be ordered, because each key is ordered on the bytes
+that converter produces — strings, chars, byte strings, every integral type, enums, floating point,
+booleans and ``DateTime`` alike. A custom key converter needs nothing added here to participate. The
+same holds for ``CborObject``, whose keys may be of any kind and may mix kinds freely within one map,
+including nulls, arrays and nested maps; that is what lets a document read off the wire be re-encoded
+deterministically whatever it contains.
 
 **An enum key whose underlying type is wider than 32 bits is written truncated to 32 bits**, and is
-therefore ordered that way too — the ordering follows the bytes actually emitted. Two enum values
-that differ only above bit 31 collide into the same map key, producing a map with duplicate keys that
-is neither deterministic nor valid to hash. This is how enums are written with or without
-``Deterministic``; if your enum is backed by ``long``/``ulong`` and its values exceed 32 bits, key
-the map by the underlying integer instead.
+therefore ordered that way too. Two enum values that differ only above bit 31 collide into the same
+map key, producing a map with duplicate keys that is neither deterministic nor valid to hash. This is
+how ``EnumConverter`` writes enums with or without ``Deterministic``; if your enum is backed by
+``long``/``ulong`` and its values exceed 32 bits, key the map by the underlying integer instead.
+
+#### The discriminator is no longer the first key
+
+Without ``Deterministic`` the discriminator is written first, at index 0. Sorted bytewise it takes
+whatever position its encoded key earns like any other: ``"_t"`` encodes to ``0x62 0x5F 0x74``, which
+places it after every one-character member name and after every two-character name starting below
+``0x5F`` — on a PascalCase model, somewhere in the middle of the map.
+
+This is what §4.2.1 requires; it grants a discriminator no exemption. Two consequences are worth
+knowing:
+
+* **Reads of polymorphic types get slower.** Deserialization bookmarks the reader, scans the map for
+  the discriminator, then rewinds and re-reads. Previously it hit on the first item; now it may scan
+  the whole map first. Correctness is unaffected — Dahomey.Cbor does not care where the discriminator
+  sits — but the cost is real.
+* **Strict readers elsewhere may reject the document.** Several polymorphic deserializers in other
+  ecosystems require the type tag in first position. A document this library round-trips happily may
+  not be readable by one of those once sorted.
+
+Both are avoidable without giving up compliance, and both are free:
+
+* ``CborObjectFormat.IntKeyMap`` — the discriminator's index is 0, encoding to the single byte
+  ``0x00``. That is the smallest encoding in the lowest major type, so nothing can sort before it.
+  This is the recommendation.
+* ``StringKeyMap`` with a one-character discriminator name, which
+  ``DefaultDiscriminatorConvention`` takes as a constructor argument. A one-character name encodes in
+  two bytes and so beats every name of two characters or more.
 
 ``Deterministic`` may be set at any point in an options object's life, including on the long-lived
 ``CborOptions.Default`` — but not while a write using those options is in flight. It is read when a
@@ -250,9 +273,13 @@ write starts, not when converters are built, so it takes effect on the very next
 *during* a write (from a property getter, a custom converter, or another thread) is picked up by the
 write after it, and the write in progress finishes on the ordering it started with.
 
-Deterministic mode also rejects any setting that would admit more than one encoding of the same
-value: setting ``ArrayLengthMode`` or ``MapLengthMode`` to ``LengthMode.IndefiniteLength`` while
-``Deterministic`` is enabled throws a ``CborException``.
+Deterministic mode also refuses any setting that would admit more than one encoding of the same
+value. An indefinite-length map or array throws a ``CborException`` at the point the header would be
+written, whichever of the three sources asked for it: ``CborOptions.ArrayLengthMode``,
+``CborOptions.MapLengthMode``, or a ``CborLengthMode`` attribute on a type or member — the last of
+which takes priority over both options, and is therefore the case most likely to surprise. The
+refusal is at the writer rather than at the property setters so that no combination can slip past by
+being configured in the wrong order.
 
 **When verifying an integrity hash, hash the bytes you received, never a re-serialization.** If
 ``UnhandledNameMode.Silent`` is in effect, decoding a document written by a newer version silently
--- a/README.md
+++ b/README.md
@@ -229,15 +229,26 @@ keys sort after all non-negative ones.
 
 Supported dictionary key types are ``string``, ``char``, ``byte[]``/``ReadOnlyMemory<byte>``, every
 integral type (``byte``, ``sbyte``, ``short``, ``ushort``, ``int``, ``uint``, ``long``, ``ulong``)
-and enums — each ordered as its own converter writes it, so an enum is ordered as text when
-``EnumFormat`` is ``WriteToString``. ``CborObject`` keys may be text strings, byte strings and
-integers, mixed freely within one map, which is what lets a document read off the wire be re-encoded
+and enums — each ordered as its own converter writes it. For an enum that means: with
+``EnumFormat.WriteToString`` a value that *has* a name is written and ordered as that name, and a
+value that has none (a combination of flags, or a cast from an unlisted number) falls back to the
+integer form in both. ``CborObject`` keys may be text strings, byte strings and integers, mixed
+freely within one map, which is what lets a document read off the wire be re-encoded
 deterministically. Any other key type throws a ``CborException`` rather than silently emitting
 unsorted output.
 
+**An enum key whose underlying type is wider than 32 bits is written truncated to 32 bits**, and is
+therefore ordered that way too — the ordering follows the bytes actually emitted. Two enum values
+that differ only above bit 31 collide into the same map key, producing a map with duplicate keys that
+is neither deterministic nor valid to hash. This is how enums are written with or without
+``Deterministic``; if your enum is backed by ``long``/``ulong`` and its values exceed 32 bits, key
+the map by the underlying integer instead.
+
 ``Deterministic`` may be set at any point in an options object's life, including on the long-lived
-``CborOptions.Default``: it is read when a value is written, not when converters are built, so it
-takes effect on the very next write.
+``CborOptions.Default`` — but not while a write using those options is in flight. It is read when a
+write starts, not when converters are built, so it takes effect on the very next write; a change made
+*during* a write (from a property getter, a custom converter, or another thread) is picked up by the
+write after it, and the write in progress finishes on the ordering it started with.
 
 Deterministic mode also rejects any setting that would admit more than one encoding of the same
 value: setting ``ArrayLengthMode`` or ``MapLengthMode`` to ``LengthMode.IndefiniteLength`` while
--- a/README.md
+++ b/README.md
@@ -219,12 +219,25 @@ deduplication meaningful. This is the §4.2.1 ordering rule, not the deprecated 
 from §4.2.3, which is not implemented.
 
 Key ordering applies to ``StringKeyMap`` and ``IntKeyMap`` objects, ``Dictionary<K,V>``, and
-``CborObject``/``CborValue`` maps; ``CborObjectFormat.Array`` has no map keys and is therefore
-already deterministic. Because ordering is on the *encoded* key, a shorter key always sorts before a
-longer one, so ``"z"`` sorts before ``"aa"``, and negative integer keys sort after all non-negative
-ones, because a negative integer is a different CBOR major type. Supported dictionary key types are
-``string``, ``int``, ``long`` and ``ulong``; any other key type throws a ``CborException`` rather
-than silently emitting unsorted output.
+``CborObject``/``CborValue`` maps. ``CborObjectFormat.Array`` writes its members positionally and has
+no map keys at all, so its bytes are identical with and without ``Deterministic``.
+
+Because ordering is on the *encoded* key, a shorter key always sorts before a longer one, so ``"z"``
+sorts before ``"aa"``. Keys of different CBOR major types order by major type first — unsigned
+integer, then negative integer, then byte string, then text string — which is why negative integer
+keys sort after all non-negative ones.
+
+Supported dictionary key types are ``string``, ``char``, ``byte[]``/``ReadOnlyMemory<byte>``, every
+integral type (``byte``, ``sbyte``, ``short``, ``ushort``, ``int``, ``uint``, ``long``, ``ulong``)
+and enums — each ordered as its own converter writes it, so an enum is ordered as text when
+``EnumFormat`` is ``WriteToString``. ``CborObject`` keys may be text strings, byte strings and
+integers, mixed freely within one map, which is what lets a document read off the wire be re-encoded
+deterministically. Any other key type throws a ``CborException`` rather than silently emitting
+unsorted output.
+
+``Deterministic`` may be set at any point in an options object's life, including on the long-lived
+``CborOptions.Default``: it is read when a value is written, not when converters are built, so it
+takes effect on the very next write.
 
 Deterministic mode also rejects any setting that would admit more than one encoding of the same
 value: setting ``ArrayLengthMode`` or ``MapLengthMode`` to ``LengthMode.IndefiniteLength`` while
--- a/README.md
+++ b/README.md
@@ -205,4 +205,33 @@ and register it the same way.
 > **Note:** the registry caches one convention per declared type, so a single hierarchy cannot mix
 > string- and int-keyed discriminators. Independent hierarchies may each use their own kind within the
 > same `CborOptions`.
+### Deterministic encoding (RFC 8949 §4.2)
+
+```csharp
+CborOptions options = new CborOptions { Deterministic = true };
+await Cbor.SerializeAsync(customObject, stream, options);
+```
+
+Guarantees the four core requirements of RFC 8949 §4.2.1: shortest-form arguments for integers,
+lengths and tags; preferred float serialization; definite lengths; and map keys sorted bytewise on
+their encoded form. The same value always produces the same bytes, which is what makes hashing and
+deduplication meaningful. This is the §4.2.1 ordering rule, not the deprecated length-first variant
+from §4.2.3, which is not implemented.
+
+Key ordering applies to ``StringKeyMap`` and ``IntKeyMap`` objects, ``Dictionary<K,V>``, and
+``CborObject``/``CborValue`` maps; ``CborObjectFormat.Array`` has no map keys and is therefore
+already deterministic. Because ordering is on the *encoded* key, a shorter key always sorts before a
+longer one, so ``"z"`` sorts before ``"aa"``, and negative integer keys sort after all non-negative
+ones, because a negative integer is a different CBOR major type. Supported dictionary key types are
+``string``, ``int``, ``long`` and ``ulong``; any other key type throws a ``CborException`` rather
+than silently emitting unsorted output.
+
+Deterministic mode also rejects any setting that would admit more than one encoding of the same
+value: setting ``ArrayLengthMode`` or ``MapLengthMode`` to ``LengthMode.IndefiniteLength`` while
+``Deterministic`` is enabled throws a ``CborException``.
+
+**When verifying an integrity hash, hash the bytes you received, never a re-serialization.** If
+``UnhandledNameMode.Silent`` is in effect, decoding a document written by a newer version silently
+drops members this version doesn't know about, and re-encoding then produces different — and
+differently hashing — bytes than what was received. Hash the wire bytes directly.
 
--- a/src/Dahomey.Cbor/CborOptions.cs
+++ b/src/Dahomey.Cbor/CborOptions.cs
@@ -51,8 +51,80 @@ namespace Dahomey.Cbor
         public DateTimeKind UnqualifiedTimeZoneDateTimeKind { get; set; }
         public CborDiscriminatorPolicy DiscriminatorPolicy { get; set; }
         public CborObjectFormat ObjectFormat { get; set; } = CborObjectFormat.StringKeyMap;
-        public LengthMode ArrayLengthMode { get; set; } = LengthMode.DefiniteLength;
-        public LengthMode MapLengthMode { get; set; } = LengthMode.DefiniteLength;
+
+        private LengthMode _arrayLengthMode = LengthMode.DefiniteLength;
+        private LengthMode _mapLengthMode = LengthMode.DefiniteLength;
+        private bool _deterministic;
+
+        public LengthMode ArrayLengthMode
+        {
+            get => _arrayLengthMode;
+            set
+            {
+                if (_deterministic)
+                {
+                    RejectIndefinite(value, nameof(ArrayLengthMode));
+                }
+
+                _arrayLengthMode = value;
+            }
+        }
+
+        public LengthMode MapLengthMode
+        {
+            get => _mapLengthMode;
+            set
+            {
+                if (_deterministic)
+                {
+                    RejectIndefinite(value, nameof(MapLengthMode));
+                }
+
+                _mapLengthMode = value;
+            }
+        }
+
+        /// <summary>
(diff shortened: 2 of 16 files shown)


src/Dahomey.Cbor/CborOptions.cs
```diff
@@ -51,8 +51,80 @@ public class CborOptions
...
+                _deterministic = value;
+            }
+        }
+
+        /// <summary>
+        /// Rejects a length mode that admits more than one encoding of the same value. Callers decide
+        /// whether deterministic mode is in force; this method does not consult <c>_deterministic</c>,
+        /// because the property setter calls it while that field is still stale.
+        /// </summary>
+        private static void RejectIndefinite(LengthMode lengthMode, string propertyName)
```
mcatanzariti, line 118: **Blocking.** This guard is in the wrong place, and it does not close the hole it is meant to close.

`CborOptions.MapLengthMode` / `ArrayLengthMode` are only the *last* resort in the length-mode resolution. Every write site resolves it as:

```csharp
lengthMode = lengthMode != LengthMode.Default
    ? lengthMode : _objectMapping.LengthMode != LengthMode.Default
        ? _objectMapping.LengthMode : _options.MapLengthMode
```

and both of the higher-priority sources are user-settable:

- `CborLengthModeAttribute` on a type → `DefaultObjectMappingConvention.cs:67` → `objectMapping.SetLengthMode(...)`
- `CborLengthModeAttribute` on a member → `DefaultObjectMappingConvention.cs:245` → `memberMapping.SetLengthMode(...)`

So `[CborLengthMode(LengthMode.IndefiniteLength)]` on a class or a property emits indefinite lengths with `Deterministic = true`, silently, and the options setters never see it. The mode is announced as guaranteeing definite lengths; right now it guarantees them only for the members nobody annotated.

Please move the check to where the effective length mode is known — either at the resolution sites, or in `CborWriter.WriteBeginMap` / `WriteBeginArray`, which is the single choke point and would also cover any converter that passes `IndefiniteLength` explicitly. That has a second benefit: throwing from a property setter on `CborOptions.Default`, which is a process-wide singleton, means an innocuous assignment in one part of an application can start throwing because another part enabled the flag. Validation at write time avoids the action-at-a-distance entirely.

Needs a test: a type with `[CborLengthMode(LengthMode.IndefiniteLength)]` serialized under `Deterministic = true`.

---

src/Dahomey.Cbor/Serialization/Converters/DeterministicKeyOrder.cs
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+            if (TryGetIntegerArgument(key, out bool negative, out ulong argument))
+            {
+                return SortKey.Integer(negative, argument);
+            }
+
+            throw new CborException(
+                $"Deterministic encoding does not define an order for {key.GetType()} map keys; "
+                + "supported key types are string, char, byte string, any integral type and enums.");
```
mcatanzariti: **Blocking.** This method decides the order by *re-implementing* what each key converter writes, rather than by asking the converter. The doc comment states the invariant correctly — "the order computed here is the order of the bytes that actually get written" — but nothing enforces it, and there are four ways it already breaks.

**1. Key types that work today start throwing.** `PrimitiveConverterProvider` hands out converters for `DateTime`, `decimal`, `bool`, `float`, `double` and `CborValue`, and `AbstractDictionaryConverter` writes keys through `KeyConverter = Registry.ConverterRegistry.Lookup<TK>()`. So `Dictionary<DateTime, T>` serializes fine on `master` and throws here the moment the flag is on. `DateTime` keys are not an exotic case. The README documents the restriction, but documenting it does not make it acceptable: turning on a serialization option should not make previously serializable models unserializable.

**2. A registered custom converter makes the order wrong, silently.** Anyone who registers their own converter for `string` or `int` — which the library explicitly supports — gets a map sorted by what this switch assumes was written, not by what was written. No exception, just a document that fails to hash the same as the equivalent document from another encoder.

**3 and 4** are in the enum branch, see the separate comment below.

The fix that removes all four at once is to encode each key once with the real `KeyConverter` into a scratch buffer and sort on those bytes. That is what §4.2.1 actually specifies — bytewise on the encoded key — so it is also a more faithful implementation than a type switch. It costs a buffer per key, but the current code already allocates a `byte[]` per text key via `Encoding.UTF8.GetBytes`, so the delta is small, and the decorate-once structure you already built is exactly the right place to hang it. It would also let you drop the "float, bool, null, array and map keys throw" limitation, since those all encode fine.

If you would rather keep the type switch, then at minimum the mode has to refuse to run when a key type resolves to a converter it does not recognise, rather than assuming the switch is exhaustive — but I would much prefer the encode-the-key approach.

---

src/Dahomey.Cbor/Serialization/Converters/DeterministicKeyOrder.cs
```diff
@@ -0,0 +1,267 @@
...
+        {
+            if (options.EnumFormat == ValueFormat.WriteToString)
+            {
+                string? name = Enum.GetName(key.GetType(), key);
+
+                if (name != null)
+                {
+                    // EnumConverter holds its names as ASCII, and an enum member name is an identifier,
+                    // so ASCII and UTF-8 agree byte for byte here.
+                    return SortKey.Text(Encoding.ASCII.GetBytes(name));
```
mcatanzariti: Two concrete instances of the mirroring problem, both of which produce an *unsorted* map with no error.

**Aliased enum members.** `Enum.GetName` returns the first name it finds for a value. `EnumConverter` builds `values2Names` in a loop over `Enum.GetNames` and overwrites on collision (`EnumConverter.cs:26-33`), so it keeps the *last*. For

```csharp
enum E { A = 1, B = 1 }
```

the sort key is `"A"` and the emitted key is `"B"`. Any other key sorting between them puts the map out of order.

**Enums narrower than 32 bits.** `EnumConverter.WriteInt32` does `Unsafe.As<T, int>(ref value)` — a 4-byte read over an enum whose storage may be 1 or 2 bytes. This code computes `(int)(uint)ToInt64(...)` instead, which sign-extends properly. For an `enum : sbyte` holding a negative value the two do not agree. You reproduce the >32-bit truncation faithfully (and document it), but not this one.

Both disappear if the sort key comes from actually running `EnumConverter` over the key, per the comment above.

Whatever route you take, please add a test for the aliased-enum case — it is the kind of thing that will otherwise be found by a user comparing hashes across two builds.

---

README.md
```diff
@@ -126,3 +126,57 @@ The last two options are useful when you write a custom cbor converter for a cla
...
+await Cbor.SerializeAsync(customObject, stream, options);
+```
+
+Guarantees the four core requirements of RFC 8949 §4.2.1: shortest-form arguments for integers,
+lengths and tags; preferred float serialization; definite lengths; and map keys sorted bytewise on
+their encoded form. The same value always produces the same bytes, which is what makes hashing and
+deduplication meaningful. This is the §4.2.1 ordering rule, not the deprecated length-first variant
+from §4.2.3, which is not implemented.
+
+Key ordering applies to ``StringKeyMap`` and ``IntKeyMap`` objects, ``Dictionary<K,V>``, and
```
mcatanzariti, line 221: **Blocking, documentation only.** The most visible behavioural change of this PR is not mentioned anywhere: **the discriminator stops being the first key of the map.**

`ObjectMapping.cs:86` inserts the discriminator member at index 0, so on `master` `_t` is always written first. Sorted bytewise, `"_t"` encodes to `0x62 0x5F 0x74`, which puts it after every one-character member name and after every two-character name starting below `0x5F` — i.e. after every PascalCase member. On a typical model it lands in the middle of the map.

This is correct behaviour: §4.2.1 fixes the order by the key bytes and grants no exemption to a discriminator, and a user who asks for deterministic output accepts what it implies. I am not asking for a benchmark and not asking you to preserve the old position. But it has to be stated, because two consequences are not obvious:

- **Reads get slower for polymorphic types.** `FindItem` (`ObjectConverter.cs:700-713`) bookmarks, scans the map for `_t`, then rewinds and re-reads. Previously it hit at the first item; now it may scan the whole map first. Correctness is unaffected — Dahomey does not care where the discriminator is — but the cost is real and worth a sentence.
- **Strict streaming readers elsewhere may reject the document.** Several polymorphic deserializers require the type tag first (System.Text.Json requires `$type` in first position, for one). A document Dahomey happily round-trips may not be readable by them once sorted.

Please also document the two ways to get the discriminator first *and* stay compliant, since both are free:

- `CborObjectFormat.IntKeyMap`: the discriminator's index is 0 (`MemberConverter.cs:400`), encoding to `0x00` — the smallest possible key encoding, in the lowest major type. Nothing can sort before it, ever. This is the recommendation to give.
- `StringKeyMap` with a chosen discriminator name: `DefaultDiscriminatorConvention` takes the name as a constructor argument, and a one-character name encodes in 2 bytes, so it beats every name of two characters or more.

---

src/Dahomey.Cbor/Serialization/Converters/ObjectConverter.cs
```diff
@@ -70,7 +87,50 @@ public struct WriterContext
...
+        /// <see cref="CborObjectFormat.Array"/> is excluded because it writes members positionally and
+        /// emits no keys at all. There is nothing to order, and reordering would move values between
+        /// array positions -- changing what the document means rather than only how it is spelled.
+        /// </para>
+        /// </remarks>
+        public IReadOnlyList<IMemberConverter> MemberConvertersForWrite
+        {
+            get
+            {
+                if (!_options.Deterministic || _objectMapping.ObjectFormat == CborObjectFormat.Array)
```
mcatanzariti, line 128: Non-blocking, but worth tightening while you are here.

This getter reads `_objectMapping.ObjectFormat` of the converter it runs on. In the polymorphic path, `Write` sets `context.objectConverter` to the *derived* type's converter and then reads `MemberConvertersForWrite` from it, while the `switch` that decides map-vs-array — and the one inside `WriteItem` that decides whether to emit a key — both read the *declared* type's `_objectMapping.ObjectFormat`.

If a base and a derived type ever disagree on `ObjectFormat`, the exclusion is evaluated against one mapping and the write against the other: members get written positionally, in sorted order, which moves values between array slots. That configuration is already broken on `master` for other reasons, so this is not a regression — but since the snapshot line below is precisely about pinning down which converter the write runs on, deciding the format from the same converter would close it for good.

---

src/Dahomey.Cbor/Serialization/Converters/AbstractDictionaryConverter.cs
```diff
@@ -57,14 +59,31 @@ public override void Write(ref CborWriter writer, TC value, LengthMode lengthMod
...
 
+        // The key set is only known at write time (unlike fixed object members, which ObjectConverter
+        // sorts once per type), so this materialises and sorts on every write. GetMapSize/WriteMapItem
+        // are untouched: they only ever pump context.enumerator, so handing them the sorted sequence's
+        // enumerator instead of the dictionary's is enough.
+        private IEnumerator<KeyValuePair<TK, TV>> SortedEntries(IDictionary<TK, TV> dictionary)
+        {
+            // Held in a field rather than written inline: the lambda closes over _options, so writing it
+            // at the call site would allocate a closure on every write.
+            _decorateKey ??= key => DeterministicKeyOrder.ForDictionaryKey(key, _options);
```
mcatanzariti: Non-blocking. Caching the delegate to avoid a closure per write is the right instinct, but `ForDictionaryKey` takes `object`, so every value-type key is boxed — one allocation per entry, per write, for `Dictionary<int, T>` and friends. The PR description's "steady-state allocation is unchanged on both paths" holds for the non-deterministic path and for reference-typed keys, not here.

`TK` is known at converter construction, so the decoration strategy can be chosen once per closed generic type instead of switching on a boxed `object` per key. That also fits naturally with encoding keys through `KeyConverter` — `ICborConverter<TK>` is already typed.

---

mcatanzariti (changes requested):
Thank you for this, and for the write-up that came with it — the quality of the reasoning in the comments and the README is well above what I usually get, and the design decisions are the right ones. The §4.2.1 reading is correct: bytewise on the *encoded* key, argument size tiered before content, major type before everything, and explicitly not the deprecated §4.2.3 length-first variant. Deciding the order once per write instead of once per comparison, and once per type for `ObjectConverter`, is the right shape. Snapshotting `MemberConvertersForWrite` into the writer context is a genuine robustness improvement independent of this feature. The observation that `CborObjectFormat.Array` has no keys and must be left alone is correct and easy to get wrong.

I can't merge it as it stands. Two blocking issues and one documentation gap, detailed inline.

## Blocking

**1. The guarantee is bypassable.** The `IndefiniteLength` refusal lives in the `CborOptions` setters, but the options are only the lowest-priority source of the length mode. `CborLengthModeAttribute` on a type or a member wins over them (`DefaultObjectMappingConvention.cs:67` and `:245`), so `[CborLengthMode(LengthMode.IndefiniteLength)]` produces indefinite lengths under `Deterministic = true` without a word. The check belongs where the effective length mode is resolved, or in `CborWriter.WriteBeginMap`/`WriteBeginArray` — one choke point that also covers converters passing the mode explicitly. See the comment on `CborOptions.cs`.

**2. Key ordering re-implements the converters instead of using them.** `DeterministicKeyOrder.ForDictionaryKey` switches on the CLR type and reproduces what it believes the key converter emits. The doc comment states the invariant — the order must be the order of the bytes actually written — and nothing enforces it. It already fails in four ways: `Dictionary<DateTime, T>` and other key types that serialize fine today throw once the flag is on; a user-registered converter for `string` or `int` yields a map sorted by an assumption rather than by the bytes, silently; aliased enum members take their sort key from `Enum.GetName` (first name) while `EnumConverter` writes `values2Names` (last name); and enums narrower than 32 bits diverge from `Unsafe.As<T, int>`. Encoding each key once through the real `KeyConverter` and sorting on those bytes removes all four, is closer to what §4.2.1 literally says, and would let you drop the float/bool/array key limitation as a bonus. See the comments on `DeterministicKeyOrder.cs`.

**3. The discriminator stops being the first key, and that is undocumented.** This is the change most likely to surprise someone. It is correct — §4.2.1 grants no exemption, and a user who asks for deterministic output accepts the consequences, so I am not asking you to preserve the old position, nor for any benchmark. But the README has to say it, along with the slower polymorphic reads it causes (`FindItem` now scans before rewinding) and the interop risk with readers that require the type tag first. And it should point at the two ways to have both, since both are free: `IntKeyMap`, where the discriminator's index 0 encodes to `0x00` and can never be outsorted, or a one-character discriminator name in `StringKeyMap`. See the comment on `README.md`.

## Non-blocking

- **Public API surface.** You offered to collapse `CborKeyComparer`; yes, please. I would make it `internal` outright — it exposes a (major type, argument, content) shape that is an artifact of the current approach, and issue 2 above may well change that shape. Nothing outside the library needs to call it.
- **Boxing per key** in the dictionary path, and the `Array`-format check reading a different mapping than the write path does. Both inline.

## Tests

87 tests is a serious effort and the tier boundaries, `int.MinValue`, `ulong.MaxValue` and mixed-kind `CborObject` cases are exactly the right things to pin down. Missing:

- `[CborLengthMode(LengthMode.IndefiniteLength)]` under `Deterministic` — the hole in issue 1
- aliased enum members — the mismatch in issue 2
- `char` and `ReadOnlyMemory<byte>` keys, both promised by the README; only `byte[]` is covered
- a nested map: every current test sorts a top-level map only
- the byte-identical-to-`master` corpus described in the PR body does not appear to be in the tree. That is the single most valuable regression test here — it is what protects everyone who never turns the flag on. Please commit it.

## Housekeeping

The branch is four commits behind `master` (#152, #153, #154, #155). I checked the merge: only `README.md` conflicts — both sides append a section — and `CborOptions.cs` merges cleanly despite `MaxDepth` landing in it. A rebase should be quick.

Happy to look again once the two blockers are addressed.

---

src/Dahomey.Cbor/Serialization/Converters/DeterministicKeyOrder.cs
mcatanzariti: You couldn't reproduce it because it doesn't reproduce — withdraw that half, don't spend time on it.

`Unsafe.As<T, int>(ref value)` forces the enum into a 4-byte stack slot, and the JIT normalises narrow integers by sign-extension when it fills that slot. So an `enum : sbyte` holding `-1` reads back as `-1`, which is exactly what `(int)(uint)ToInt64(...)` computes. The two agree. I asserted a divergence from the shape of the code without a case behind it.

The aliased-enum half of this comment stands, and your correction comment gets it right: `Enum.GetName` first-name versus `values2Names` last-name, sorting a 5-byte key and writing a 9-byte one. Thank you for going back and checking the non-aliased case rather than leaving the larger claim standing — that saved us both a wrong turn.

Neither changes the fix. Encoding through the real `KeyConverter` removes the alias case for the right reason, and it would have removed the narrow-enum case too had it been real.

---

README.md
```diff
@@ -205,4 +205,57 @@ and register it the same way.
...
+``CborObject``/``CborValue`` maps. ``CborObjectFormat.Array`` writes its members positionally and has
+no map keys at all, so its bytes are identical with and without ``Deterministic``.
+
+Because ordering is on the *encoded* key, a shorter key always sorts before a longer one, so ``"z"``
+sorts before ``"aa"``. Keys of different CBOR major types order by major type first — unsigned
+integer, then negative integer, then byte string, then text string — which is why negative integer
+keys sort after all non-negative ones.
+
+Supported dictionary key types are ``string``, ``char``, ``byte[]``/``ReadOnlyMemory<byte>``, every
```
mcatanzariti: **Blocking.** This paragraph now describes code that no longer exists. `b01cbc8` widened key support without the documentation following, so three statements in this section are false as written:

1. **"Any other key type throws a `CborException`"** — no longer true. Keys go through their own converter now, so `DateTime`, floating point and booleans work. Your own commit message says so: *"floating point dictionary keys and a `CborObject` with a boolean key both sort correctly instead of throwing."* The enumerated list of supported types is the wrong shape entirely now — the rule is "any key type whose converter can write it", which is shorter and true.

2. **The enum paragraph** (line 240 onwards) explains ordering in terms of `Enum.GetName` and the `WriteToString` fallback. That mechanism is gone; ordering follows `EnumConverter`'s output because it *is* `EnumConverter`'s output. The >32-bit truncation warning stays true and worth keeping — it is a property of `EnumConverter`, not of the ordering — but the surrounding explanation needs rewriting.

3. **"Deterministic mode also rejects any setting…"** (line 253) mentions only `ArrayLengthMode` and `MapLengthMode`. Since `a5b3fa8` the refusal is at the writer and covers `CborLengthMode` on a type or member, which outranks the options. That is the case a reader most needs to know about, and it is the one the sentence omits.

The PR description carries the same staleness — "Known limitations" still says float, bool, array and map keys throw, and "Public API" no longer matches. Worth a pass before merge so the description does not become the commit message of record for behaviour that changed.

---

src/Dahomey.Cbor/Serialization/CborKeyComparer.cs
```diff
@@ -0,0 +1,160 @@
...
+        /// </param>
+        /// <param name="contentA">
+        /// Key A's raw payload when <paramref name="majorTypeA"/> is a string type: UTF-8 bytes for
+        /// text, the bytes themselves for a byte string, in both cases without the CBOR header.
+        /// Ignored for integer types.
+        /// </param>
+        /// <param name="majorTypeB">Key B's CBOR major type.</param>
+        /// <param name="argumentB">Key B's CBOR argument, as above.</param>
+        /// <param name="contentB">Key B's raw payload, as above.</param>
+        public static int CompareKeys(
```
mcatanzariti: The rewrite left this class half dead, and it is still `public`.

After `b01cbc8`, `CompareKeys` and `CompareIntegerKeys` have no caller in `src/Dahomey.Cbor` at all — the only references are in `DeterministicEncodingTests`. They existed to serve `DeterministicKeyOrder`'s `SortKey`, which no longer exists. Please delete both, along with the tests that exercise them directly; what those tests pin is now pinned by the encoded-bytes comparison itself.

`CompareIntKeys` and `CompareTextKeys` are still live, called from `ObjectConverter.CompareMembersForDeterministicOrder`. That path is fine and does not need the converter treatment: `MemberName` is the exact byte span handed to `WriteString`, and `MemberIndex` the exact value handed to `WriteInt32`, so there is no prediction involved — the comparison is over the bytes that get written.

With those two gone, please make the class `internal` as agreed. It exposes a (major type, argument, content) shape that was an artifact of the old approach, and keeping it public would freeze a signature the library no longer uses.

---

mcatanzariti (changes requested):
Both code blockers are gone, and you fixed them the right way rather than the quick way. Rebase is clean, CI is green on `b01cbc8`. I've resolved the four threads that this covers.

**Blocker 1 (`a5b3fa8`).** The refusal is at `WriteBeginMap`/`WriteBeginArray`, wired through the four `Cbor.cs` construction sites, with the setter checks kept as a configuration-time convenience and the writer as the backstop. Naming all three sources in the exception message is the right instinct — "indefinite lengths are refused" would have sent people looking at options they never set. `DeterministicLengthModeTests` pins it from both directions, including `TheAttributeStillWorksWithoutDeterministic`, which is the test I would have asked for if it weren't already there.

**Blocker 2 (`b01cbc8`).** This is the rewrite, not a patch over the symptoms. Keys encoded once through their own converter into a single buffer, offsets recorded, an index permutation sorted on the encoded spans. The whole class of defect is gone by construction — aliased enums, custom converters, `DateTime`, floats, booleans — and the boxing per key went with it, so that non-blocking point is closed as a side effect. Dropping the separate length comparison because the length already lives in the header byte where the comparison starts makes the code both shorter and more literally §4.2.1 than what it replaced. The commit message is honest about the second encoding at write time, which is the right trade to have made.

Two things left, neither large.

**Blocker 3 is untouched, and the fixes have now made the README wrong.** No commit since 2026-08-02 touches `README.md`. The discriminator no longer being the first key is still undocumented — that was the original ask — and on top of it, widening key support in `b01cbc8` left three statements in that section describing code that no longer exists. Details inline. This is the only thing standing between the PR and merge, and it is all prose.

**`CborKeyComparer` is half dead and still public.** `CompareKeys` and `CompareIntegerKeys` lost their last production caller in the rewrite; only the tests reach them now. Delete those two, then make the class `internal` as you agreed. Inline.

Also still open from last round, both small and both previously accepted: the `Array`-format check in `ObjectConverter` reading the derived mapping while the write path reads the declared one, and the byte-identical-to-`master` corpus. The corpus matters more than it did before — the key path was rewritten wholesale this round, and that corpus is what would prove the `Deterministic = false` output did not move. A nested-map test would be worth having too, since every current ordering test is on a top-level map.

The per-key-type test matrix I asked for last time I would no longer insist on. The mechanism is uniform now, so a `char`-keyed dictionary tests the same code path as every other key; the three targeted tests you added (alias, custom converter, `DateTime`) are the ones that carry real information.

Send the docs and the `CborKeyComparer` cleanup and I'll merge.

---

mcatanzariti (approved):
Approved. Everything from both rounds is in, and I've resolved the remaining threads.

`ca4b1c2` closes the last three. The README now says the discriminator is no longer the first key, gives both consequences — the scan-then-rewind cost on polymorphic reads, and strict readers elsewhere that want the type tag first — and puts the two free workarounds next to them with `IntKeyMap` as the recommendation. That is exactly the shape I wanted: the consequence stated where someone enabling the flag will actually meet it, not buried. The three statements the key rewrite had falsified are corrected, and the PR description was brought back in line too.

`CborKeyComparer` is `internal` with `CompareKeys` gone and `CompareIntegerKeys` folded into `CompareIntKeys`, which still handles `int.MinValue` correctly — `-1L - a` stays in `long`. On the strong-naming: I derived the public key blob from `src/Dahomey.snk` and it matches the `InternalsVisibleTo` value byte for byte. The library was already strong-named with that key, so signing the test project is the ordinary consequence rather than anything new, and the comments in both csproj files explain why to whoever finds them next.

The object format is now snapshotted from the same converter as the member list, with all three write-side switches reading it.

`DeterministicByteCompatibilityTests` is the part I'm most glad to have. Seventeen entries, insertion orders deliberately unsorted, and an assertion that fails if a corpus entry has no pinned bytes — so a case cannot be added and left unpinned. I checked several of the expected values by hand: `object.stringKeyMap` is in declaration order, `object.intKeyMap` gives `-1, 0, 7` unsorted, `object.polymorphic` has `_t` first, and `dictionary.double` is correct binary16 for both values. Those are captured bytes, not predicted ones, and they are what protects every caller who never turns the flag on.

Two notes for later, neither affecting this PR.

The read paths still take `ObjectFormat` from the declared type's mapping while the write now takes it from the derived converter. For a base and derived that disagree on format the asymmetry simply points the other way than it did; the permanent fix is to reject disagreeing formats when the mapping is built, which is its own issue rather than yours here.

On merge order: this goes in first. I simulated it — #161 merges onto master+this with no conflict at all, #156 and #157 hit only positional collisions in `Dahomey.Cbor.csproj` and `CborOptions.cs`. #156 in particular is semantically clear, since its generator emits `new ObjectConverter<T>(options, factory)` and so inherits deterministic member ordering rather than bypassing it. #159 is the one to sequence last, and I've left a note there.

Thank you for taking the rewrite in the second round instead of patching the four symptoms, and for going back to correct your own overstatement about the string-enum case. Both made this shorter than it would have been.

---

PR #159, opened 2026-08-02
RFC 8610 CDDL schema generation from [CborSerializable] types

Rafael-SOWNet:
> **Stacked on #156 — please review that one first.**
> This branch is cut from `aot`, not `master`, because it reads the `[CborSerializable]` contexts that #156 introduces. GitHub therefore shows #156's diff underneath this one. **The CDDL work is the last 17 commits, `a3c1500..104a681`**; everything before that belongs to #156. This cannot merge before #156 does, and I'm happy to rebase or resubmit however suits you.

## What this adds

`[CborCddlSchema]` on a source-generated context emits an RFC 8610 CDDL schema for every type the context declares, as a `const string`:

```csharp
[CborSerializable(typeof(Person))]
[CborCddlSchema]
public partial class MyContext : CborSerializerContext { }

File.WriteAllText("person.cddl", MyContext.CddlSchema);
```

The point is that a wire-format change becomes a reviewable diff rather than a field mismatch in someone's production system, and a non-.NET consumer gets a machine-checkable contract.

It is a second emit target inside the existing generator, opt-in per context. A context without the attribute generates exactly what it generates today — covered by a test.

## The governing rule

**The schema describes what the serializer *writes* — exact and closed.** It is not a description of what the reader tolerates; `UnhandledNameMode` defaults to `Silent`, so unknown keys are accepted on read but never emitted. That choice is what makes the round-trip test meaningful: the serializer's own output must validate against its own schema.

Output uses core RFC 8610 grammar only — prelude types, arrays, maps, `/`, ranges, `#6.n(...)`. No control operators.

## What it covers

- All three `CborObjectFormat` values, with `IntKeyMap`/`Array` members emitted in the ascending-index order `ObjectMapping` actually re-sorts them into at registration — for `Array` that is positional correctness, not tidiness.
- Polymorphism, including the discriminator that `SetDiscriminator` inserts *after* `SetMemberMappings` and which is therefore absent from the generator's member list. Because the effective default policy is `Auto`, presence depends on the static type at the call site, so each subtype gets two rules: `X` when written as itself, `X-poly` when written through a base. String and int discriminators, all three placements (`"_t"`, key `0`, and the tag-wrapped first array element).
- RFC 8746 typed arrays, enums including `[Flags]`, nullability taken from nullable-reference annotations, collections, dictionaries, `byte[]`, and `DateTime` in all three `DateTimeFormat` shapes.
- Interfaces as polymorphic bases, not just base classes.

A type with no CDDL representation is a **build error** (`CBOR1007`), never a silent omission — a quietly incomplete schema is worse than no schema.

## Three new `[CborSourceGenerationOptions]` properties

`EnumFormat`, `DateTimeFormat` and `TypedArrayMode` are now declarable on a context. Each changes the bytes and none was previously expressible, so a generated context could not express enum-as-string at all. This is a fix to the AOT path in its own right, independently tested, and the schema needs it because the generator cannot see run-time configuration.

## Validation

There is no CDDL implementation for .NET, so emitted schemas are checked against the Ruby `cddl` gem — the reference implementation. CI installs it and sets `CDDL_REQUIRED=1`, under which a missing gem fails the build rather than skipping, so the check cannot silently lapse.

Three layers: schemas are pinned as exact text so any change is a reviewable diff; every emitted schema is run through the gem's parser; and real serializer output is validated against it, with negative cases that must be rejected.

**1145 tests pass on net10.0**, 0 skipped. All four TFMs build.

Some mappings are measured rather than derived. `double` and `float` map to the prelude `float`, not `float64`/`float32`, because `CborWriter.WriteDouble` shrinks to the shortest round-tripping form in *every* mode — `1.5` goes out as `f9 3e00`, and `float64` rejects it. Likewise `[Flags]` enums get the open integer form, because `EnumConverter` writes the raw underlying value and a combination has no declared name.

## Two pre-existing bugs found while mapping types

Both are outside this PR and unfixed here; I'll file them separately if useful.

1. **`decimal` serialises to ill-formed CBOR.** `CborWriter` writes `0xFC` followed by 16 bytes, but additional information 28 is reserved under RFC 8949 §3, so no conforming decoder can read it — a standard decoder rejects it outright. Tag 4 (decimal fraction) is the interoperable encoding. This PR reports `CBOR1007` for `decimal` rather than inventing a mapping.
2. **`EnumConverter.WriteInt32` truncates `long`-backed enums.** It reinterprets via `Unsafe.As<T, int>`, so a member outside `int` range is written truncated.

`Guid`, `char`, `DateTimeOffset` and scalar `Half` have no converter at all and likewise report `CBOR1007`.

## Known limitations, stated rather than hidden

- Member types follow their nullable annotations. A member declared non-nullable but left null at run time is written as `F6` and will not validate — the annotation is taken as the developer's contract, as `System.Text.Json`'s schema exporter does.
- The discriminator key is assumed to be `_t`; a custom `DefaultDiscriminatorConvention` member name is a run-time registration the generator cannot see. The emitted header says so.
- The schema is a compile-time artifact, so mutating `CborOptions` after `Configure` invalidates it.


src/Dahomey.Cbor.Generator/CddlEmitter.cs
```diff
@@ -0,0 +1,667 @@
...
+            bool hasNamespace = !contextSymbol.ContainingNamespace.IsGlobalNamespace;
+            string indent = hasNamespace ? "    " : string.Empty;
+
+            if (hasNamespace)
+            {
+                builder.AppendLine($"namespace {contextSymbol.ContainingNamespace.ToDisplayString()}");
+                builder.AppendLine("{");
+            }
+
+            builder.AppendLine($"{indent}public partial class {contextSymbol.Name}");
```
mcatanzariti, line 652: This re-declares the context as `public partial class <SimpleName>` directly under the namespace, but `Emitter.Emit` already solves this properly — it walks the `ContainingType` chain, emits `Accessibility(contextSymbol)`, and uses `NameWithTypeParameters(contextSymbol)` (`Emitter.cs:47-63`). Three cases break here:

- **`internal` context** → the two partial declarations disagree on accessibility (CS0262).
- **Nested context** → this emits a *new* top-level `Inner` alongside the real `Outer.Inner`, so `Outer.Inner.CddlSchema` doesn't exist.
- **Generic context** → partial declarations differ in type parameters (CS0264).

Since `[CborCddlSchema]` is opt-in on a context that already works, this turns a compiling project into a build failure purely by adding the attribute. Reusing the same header-emitting helper as `Emitter` (or factoring it out) would keep the two paths from drifting.

---

src/Dahomey.Cbor.Generator/CddlTypeReference.cs
```diff
@@ -0,0 +1,308 @@
...
+                        "Unix" => "#6.1(int)",
+                        "UnixMilliseconds" => "#6.1(float)",
+                        _ => "#6.0(tstr)",
+                    };
+            }
+
+            // System.Decimal is deliberately absent: it is written as 0xFC plus 16 bytes, and
+            // additional information 28 is reserved and ill-formed under RFC 8949 section 3, so no
+            // conforming decoder can read it and no CDDL can describe it. Guid, DateTimeOffset and
+            // char have no scalar converter either. Nor does System.Half: the only place the library
```
mcatanzariti, line 269: The premise for `char` is incorrect: it *does* have a scalar converter. `PrimitiveConverterProvider.cs:22` maps `TypeCode.Char` to `CharConverter`, whose `Write` calls `CborWriter.WriteChar` (`CborWriter.cs:343`), which UTF-8 encodes the single char and writes it as a CBOR text string. `TypeCollector.IsPrimitive` also lists `SpecialType.System_Char`, so a `char` member reaches `RenderPrimitive`, finds no case, and becomes a hard `CBOR1011`.

So adding `[CborCddlSchema]` to a context whose model has a `char` member fails the build for a member that serializes fine today. `case SpecialType.System_Char: return "tstr";` covers it exactly.

Worth noting the pinned test asserting `char` is unsupported encodes the same false premise, so it won't catch this — it will need updating alongside.

(`Guid`, `DateTimeOffset` and scalar `Half` in this comment are correct — no converter for any of them.)

---

src/Dahomey.Cbor.Generator/CddlTypeReference.cs
```diff
@@ -0,0 +1,308 @@
...
+            return type is INamedTypeSymbol { TypeArguments.Length: > 0 } named
+                ? named.TypeArguments[0]
+                : model.UnderlyingType!;
+        }
+
+        /// <summary>
+        /// Narrow integers get a range rather than the prelude's unbounded <c>int</c>, because the
+        /// schema is a contract about what is written and a byte is never 256.
+        /// </summary>
+        private static string? RenderPrimitive(ITypeSymbol type, GenerationOptions options)
```
mcatanzariti, line 210: `BigInteger` falls through this method unhandled, and unlike `decimal` that looks unintended rather than decided.

`TypeCollector.IsPrimitive` explicitly opts it in (`TypeCollector.cs:611-617`), with a comment saying `PrimitiveConverterProvider` resolves it to a concrete `BigIntegerConverter` so "it belongs with the cases above". But this switch keys off `SpecialType`, and `BigInteger` has none — so it classifies as `Primitive`, renders `null`, and reports `CBOR1011`. It's also absent from the documented-exclusions comment below and has no test either way.

An explicit `type.ToDisplayString() == "System.Numerics.BigInteger"` check here would mirror the one in `TypeCollector` and keep the two in agreement. The natural rendering is the tag 2 / tag 3 bignum pair (`#6.2(bstr) / #6.3(bstr)`), narrowed to whatever `BigIntegerConverter` actually emits for small values — worth checking the converter before pinning it.

---

src/Dahomey.Cbor.Generator/CddlEmitter.cs
```diff
@@ -0,0 +1,667 @@
...
+            }
+
+            builder.Append(ruleNames[key]);
+            builder.Append(" = ");
+
+            bool isFlags = model.Symbol.GetAttributes().Any(attribute =>
+                attribute.AttributeClass?.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat)
+                    == "global::System.FlagsAttribute");
+
+            if (isFlags)
```
mcatanzariti, line 498: The `values.Count == 0` fallback below guards only the non-flags path. A `[Flags]` enum with no declared members under `EnumFormat = WriteToString` makes `string.Join(" / ", names)` return `""`, so this emits:

```cddl
E =  / uint
```

which is not parseable CDDL — meaning it's the one output shape the `cddl` gem validation can't help with, since the schema fails at the parse step. Hoisting the emptiness check above the `isFlags` branch (falling back to the bare `integerForm`) fixes it.

Narrow case, but a memberless `[Flags]` enum is legal C# and costs one line to handle.

---

src/Dahomey.Cbor.Generator/TypeNames.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// <c>N-Outer-Inner</c>: the type's namespace (dots flattened to '-', omitted when global),
+        /// then its containing types outermost-first, then its own name. RFC 8610 ids accept '-'
+        /// between alphanumerics, so the flattened chain is a legal rule name fragment. Unlike
+        /// <see cref="AccessorName"/>, this reaches the containing-type chain too, so nested
+        /// <c>N.Outer.Inner</c> and <c>N.Other.Inner</c> -- indistinguishable by <c>type.Name</c> alone
+        /// -- come out as <c>N-Outer-Inner</c> and <c>N-Other-Inner</c>.
+        /// </summary>
+        private static string QualifiedSimpleName(ITypeSymbol type)
```
mcatanzariti, line 126: Minor, and only reachable on the collision-resolution path: namespace separators and containing-type separators are both flattened to `-`, so the mapping isn't injective. Namespace `N.A` + type `B` and namespace `N` + nested type `A.B` both come out as `N-A-B`.

Since this method is precisely what's used to *break* a collision, two colliding types that flatten to the same string would emit duplicate rule names — a redefinition the schema can't recover from. Using a different separator for the namespace/type boundary (or appending a disambiguating suffix when the resolved names still collide) would close it.

Low likelihood in practice; flagging it because the doc comment claims the qualified form "never collides again".

---

src/Dahomey.Cbor.Generator/TypeNames.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+            if (type is INamedTypeSymbol { IsGenericType: true } named)
+            {
+                string baseName = named.Name;
+                string arguments = string.Concat(named.TypeArguments.Select(AccessorName));
+                return $"{baseName}Of{arguments}";
+            }
+
+            return type.Name;
```
mcatanzariti, line 49: Rule names are the raw C# type name, but RFC 8610's `id` production is ASCII-only (`ALPHA *(... ALPHA / DIGIT)`), while C# identifiers allow any Unicode letter. A type named e.g. `Café` or with Cyrillic/CJK characters emits a rule name the gem's parser rejects, so the whole schema fails to parse.

Not worth a full transliteration scheme — sanitizing non-ASCII characters to `_` (or reporting a diagnostic pointing at the type) would be enough to fail loudly at the right place instead of producing an unparseable artifact.

Low likelihood, but cheap to guard.

---

mcatanzariti (commented):
Reviewed the CDDL emitter, the three new `[CborSourceGenerationOptions]` properties and their tests, checking the emitter's claims against the runtime library rather than against the PR description.

What holds up: the RFC 8746 tag table matches `TypedArrayTags`; the discriminator placement and `Auto` policy resolution match `ObjectMapping.SetDiscriminator` / `DiscriminatorMemberConverter.ShouldSerialize`, and the two-rule `X` / `X-poly` split is the right consequence of that; `float` rather than `float64`/`float32` is correct, since `CborWriter.WriteDouble` shrinks to the shortest round-tripping form unconditionally; `#6.0(tstr)` / `#6.1(int)` / `#6.1(float)` match `DateTimeConverter`; and the ascending-index re-sort matches what `ObjectMapping` does at registration, which for `Array` is positional correctness rather than tidiness. The `decimal` exclusion is right — additional information 28 really is reserved under RFC 8949 §3. Validating against the Ruby `cddl` gem with `CDDL_REQUIRED=1` so a missing gem fails rather than skips is the right call.

Six inline comments. The first is the one I'd consider blocking:

1. **`CddlEmitter.EmitSource` re-declares the context header itself** instead of reusing the logic in `Emitter.Emit`, dropping accessibility, the containing-type chain and type parameters. An `internal`, nested or generic context stops compiling as soon as `[CborCddlSchema]` is added.
2. **`char` is treated as having no converter, but it does** — `CharConverter` writes a `tstr`. An existing `char` member becomes a `CBOR1011` build error. The test pinning this asserts the same false premise.
3. **`BigInteger` is opted in by `TypeCollector.IsPrimitive` but has no `RenderPrimitive` case**, so it errors out. Having now read `CborWriter.WriteBigInteger`, the exact shape is `int / #6.2(bstr) / #6.3(bstr)`: values within ±`ulong.MaxValue` go out as basic integers and only larger magnitudes reach the bignum tags.
4. **A memberless `[Flags]` enum under `EnumFormat = WriteToString` emits `E =  / uint`**, which no CDDL parser accepts — one line to guard.
5. `QualifiedSimpleName` flattens namespace and containing-type chains with the same `-`, so the collision-breaking name can itself collide. Low likelihood.
6. Rule names are raw C# type names, but RFC 8610's `id` is ASCII-only, so a non-ASCII type name emits an unparseable rule. Low likelihood.

Also checked and found fine: context-level `DiscriminatorPolicy` does reach `model.DiscriminatorPolicy`; `CborIndex!.Value` cannot be null, since `TypeCollector` drops indexless members for non-`StringKeyMap` formats; `ElementOf`/`ValueOf` are safe because `Classify` requires exact arity; `AnyDiscriminatorBeneath` is acyclic; inherited members are collected via `EnumerateMembers`; and the `?` suffix in `MemberTypeName` doesn't affect converter resolution.

The stated limitations — nullable annotations taken as the contract, the assumed `_t` discriminator key, the compile-time nature of the artifact — are the right ones to document rather than paper over.

---

Issue #149, opened 2026-03-10
Stack Overflow


troplin:
The following code leads to a stack overflow:

```cs
class TestObject
{
    public TestObject? TestProperty { get; set; } = null;
}
```

```cs
var buffer = new ArrayBufferWriter<byte>();
Cbor.Serialize(new TestObject(), buffer);
```

PR #160, opened 2026-08-04
Build converters for mutually referencing types without overflowing the stack

Rafael-SOWNet:
Two types that reference each other cannot be serialized at all today — the process dies on an uncatchable `StackOverflowException` before any CBOR is written:

```csharp
public class Parent { public string Name { get; set; } public Child Only { get; set; } }
public class Child  { public int Age { get; set; }     public Parent Owner { get; set; } }

Cbor.Serialize(new Parent { Name = "root", Only = new Child { Age = 3 } }, buffer);
```

No cycle in the *data* is needed — `Only.Owner` is null above. The recursion is in building the converter graph, and a plain parent/child model with a back-reference is enough to trigger it.

## Why it happens

A converter resolves its members' converters while it is being constructed, and it is not in `CborConverterRegistry._converters` until that constructor returns. So `Lookup(Parent)` starts building `ObjectConverter<Parent>`, which looks up `Child`, which starts building `ObjectConverter<Child>`, which looks up `Parent` — not registered yet, because its constructor has not returned — and around it goes. `ConcurrentDictionary.GetOrAdd` runs its factory outside the lock, so nothing intervenes.

The existing break covers only the direct `A -> A` case: `MemberMapping<T>.InitializeConverter` reuses the parent converter when a member's type is the type being built. A back-reference through a second type is not covered, and neither is any longer cycle.

This is a different failure from the one bounded by `CborOptions.MaxDepth` in https://github.com/dahomey-technologies/Dahomey.Cbor/pull/155. That one is a write-side recursion over the *data*, on a converter graph that built fine. This one never gets as far as looking at a value.

## What this changes

* `CborConverterRegistry` records which types are being constructed on the current thread, and exposes `IsUnderConstruction`. Thread-static and keyed by `(registry, type)` so two registries do not interfere; the set is empty except while a converter is being built.
* `MemberMapping<T>.InitializeConverter` consults it. If the member's own type is already in flight, it leaves the converter unresolved rather than recursing, and the `Converter` getter completes the lookup on first use — by which time the graph is built.
* `MemberConverter<T, TM>` and `StructMemberConverter<T, TM>` resolve their converter on first use instead of in their constructor. This is the part that actually lets a cycle close: a member converter has to be able to exist before the converter it points at does.

## What it deliberately does not change

* **An actual object-graph cycle (`obj.Property = obj`) still throws.** That is a write-side recursion over the data, bounded by `MaxDepth`, and it stays a `CborException`. Pinned as a test here so the two failure modes do not get conflated later.
* **No wire-format change.** The same converters are chosen for the same types; only the moment of resolution moves.
* **Error timing for a bad `[CborConverter]` on a deferred member.** `VerifyMemberConverterType` still runs, but for a member whose resolution was deferred it runs on first use rather than during construction. This is reachable only in the cycle case — which previously could not be reached at all — so no currently-working code changes behaviour.
* **Nothing about registry thread safety.** Two threads building the same graph still each complete independently, and `GetOrAdd` settles which instance wins, exactly as before.

## Tests

Four new cases in `src/Dahomey.Cbor.Tests/Issues/Issue0149Cycles.cs`: a two-type cycle written and read back, a three-type cycle (nothing in the break may depend on the participants being adjacent), and the object-graph cycle that must still throw.

**All four crash the test host without this change.** That is also why they could not have been added alongside https://github.com/dahomey-technologies/Dahomey.Cbor/pull/152 — a `StackOverflowException` cannot be caught or asserted on, it takes the process with it. It is the same reason https://github.com/dahomey-technologies/Dahomey.Cbor/issues/149 went so long without a test for the case it was originally reported for.

Full suite: 978 passing, green on .NET 8, 9 and 10 — https://github.com/Rafael-SOWNet/Dahomey.Cbor/actions/runs/30959882309

GitHub is holding this PR's own `Build and Test` run for approval (first-time contributor), so that run is from my fork. The branch behind it, `ci-check/cycle-safe-converters`, differs from this PR's head by exactly one commit adding `workflow_dispatch:` to the workflow file; `src/` is byte-identical.

## How I found it

Writing a test for point 4 of your review on https://github.com/dahomey-technologies/Dahomey.Cbor/pull/156 — you asked for defects to be confirmed with a failing test before being acted on, which was good advice. The test did not show the silent mis-binding described there; it took the test host down instead, and the same shape turned out to fail on `master` with no context and no generator anywhere near it.

Cut from `master`, independent of every other open PR.

### DIFF
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0149Cycles.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0149Cycles.cs
@@ -1,0 +1,134 @@
+using Dahomey.Cbor.Tests.Extensions;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests.Issues
+{
+    /// <summary>
+    /// Issue #149, converter-construction side: two types that reference each other must be
+    /// serializable.
+    /// </summary>
+    /// <remarks>
+    /// A converter resolves its members' converters while it is being constructed, and it is not in
+    /// the registry until its constructor returns. Two mutually referencing types therefore ask the
+    /// registry for each other indefinitely, and the process dies on an uncatchable
+    /// <see cref="System.StackOverflowException"/> before any CBOR is written — a different failure
+    /// from the deep-nesting one <c>CborOptions.MaxDepth</c> bounds, which happens during a write on
+    /// a converter graph that built fine.
+    /// <para>
+    /// The single-type case (<c>A.Property</c> of type <c>A</c>) is covered by
+    /// <see cref="Issue0149"/>; these are the cases that need more than the parent-converter reuse.
+    /// </para>
+    /// </remarks>
+    public class Issue0149Cycles
+    {
+        public class Parent
+        {
+            public string Name { get; set; }
+            public Child Only { get; set; }
+        }
+
+        public class Child
+        {
+            public int Age { get; set; }
+            public Parent Owner { get; set; }
+        }
+
+        public class RingA
+        {
+            public int Id { get; set; }
+            public RingB Next { get; set; }
+        }
+
+        public class RingB
+        {
+            public int Id { get; set; }
+            public RingC Next { get; set; }
+        }
+
+        public class RingC
+        {
+            public int Id { get; set; }
+            public RingA Next { get; set; }
+        }
+
+        /// <summary>
+        /// Two types referencing each other — the back-reference shape that any parent/child model
+        /// has.
+        /// </summary>
+        [Fact]
+        public void MutuallyReferencingTypesCanBeSerialized()
+        {
+            Parent parent = new Parent
+            {
+                Name = "root",
+                Only = new Child { Age = 3 },
+            };
+
+            // a2                        map(2)
+            //    644e616d65             "Name"
+            //    64726f6f74             "root"
+            //    644f6e6c79             "Only"
+            //    a2                     map(2)
+            //       63416765            "Age"
+            //       03                  3
+            //       654f776e6572        "Owner"
+            //       f6                  null
+            Helper.TestWrite(parent, "A2644E616D6564726F6F74644F6E6C79A26341676503654F776E6572F6");
+        }
+
+        [Fact]
+        public void MutuallyReferencingTypesCanBeDeserialized()
+        {
+            const string hexBuffer = "A2644E616D6564726F6F74644F6E6C79A26341676503654F776E6572F6";
+
+            Parent parent = Cbor.Deserialize<Parent>(hexBuffer.HexToBytes());
+
+            Assert.Equal("root", parent.Name);
+            Assert.Equal(3, parent.Only.Age);
+            Assert.Null(parent.Only.Owner);
+        }
+
+        /// <summary>
+        /// A cycle longer than two types: nothing in the break may depend on the pair being adjacent.
+        /// </summary>
+        [Fact]
+        public void ThreeTypeCycleCanBeSerialized()
+        {
+            RingA a = new RingA
+            {
+                Id = 1,
+                Next = new RingB { Id = 2, Next = new RingC { Id = 3 } },
+            };
+
+            // a2                     map(2)
+            //    624964              "Id"
+            //    01                  1
+            //    644e657874          "Next"
+            //    a2                  map(2)
+            //       624964           "Id"
+            //       02               2
+            //       644e657874       "Next"
+            //       a2               map(2)
+            //          624964        "Id"
+            //          03            3
+            //          644e657874    "Next"
+            //          f6            null
+            Helper.TestWrite(a, "A262496401644E657874A262496402644E657874A262496403644E657874F6");
+        }
+
+        /// <summary>
+        /// An actual object graph cycle is still a write-side infinite recursion, bounded by
+        /// <see cref="CborOptions.MaxDepth"/> rather than by anything here. Building the converters
+        /// is what must not overflow; writing a cycle is a caller error and stays a
+        /// <see cref="CborException"/>.
+        /// </summary>
+        [Fact]
+        public void AnActualObjectCycleIsStillBoundedByMaxDepth()
+        {
+            Parent parent = new Parent { Name = "root" };
+            parent.Only = new Child { Age = 3, Owner = parent };
+
+            Assert.Throws<CborException>(() => Helper.Write(parent));
+        }
+    }
+}
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0149Cycles.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0149Cycles.cs
@@ -1,4 +1,6 @@
 using Dahomey.Cbor.Tests.Extensions;
+using System;
+using System.Collections.Generic;
 using Xunit;
 
 namespace Dahomey.Cbor.Tests.Issues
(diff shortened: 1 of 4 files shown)


mcatanzariti (changes requested):
Thanks for this — the diagnosis is correct and the fix is well scoped. `MemberMapping.InitializeConverter` calls `_converterRegistry.Lookup(MemberType)` while the owning converter's constructor is still running, and `GetOrAdd` runs its factory outside any lock with nothing in `_converters` until that constructor returns, so `Parent → Child → Parent` recurses until the stack is gone. The existing break only covers the direct `A → A` member, which is why the shape originally reported in #149 stayed broken. Making the member converters resolve on first use is the right lever: a member converter has to be constructible before the converter it points at exists.

Two blocking items, then some non-blocking notes.

Note that I reviewed this statically — I was not able to build and run the suite locally.

---

## Blocking

### 1. `_converterDeferred` introduces a data race, and it is redundant

In the new `Converter` getter the completion runs outside `lock (this)` — `EnsureInitialize` returns immediately once `_isInitialized` is set — and writes two non-volatile fields in this order:

```csharp
_converter = _converterRegistry.Lookup(MemberType);
VerifyMemberConverterType(_converter.GetType());
_converterDeferred = false;
```

A concurrent reader on a weak memory model (arm64) can observe `_converterDeferred == false` while `_converter` is still null, skip the block, and return null. That null then flows into `(ICborConverter<TM>)_memberMapping.Converter!` in `MemberConverter.Converter`, gets cached by the `??=`, and throws a `NullReferenceException` on the next `Read`/`Write`. This is reachable in practice: object mappings — and therefore `MemberMapping` instances — are cached per type in `ObjectMappingRegistry`, so two threads serializing `Parent` and `Child` against the same `CborOptions` share the mapping.

The flag also carries no information that `_converter` does not already carry. After `EnsureInitialize`, `_converter == null` can only mean deferral: the `[CborConverter]` branch assigns or throws, the parent-reuse branch assigns, and `Lookup` never returns null. Dropping the flag and completing under the existing lock removes the race and shrinks the change:

```csharp
public ICborConverter? Converter
{
    get
    {
        EnsureInitialize();

        // A cycle deferred this resolution during initialization. By the time anyone asks
        // for the converter the graph is built, so the lookup now completes.
        if (_converter == null)
        {
            lock (this)
            {
                _converter ??= _converterRegistry.Lookup(MemberType);
            }
        }

        return _converter;
    }
}
```

### 2. `VerifyMemberConverterType` on the deferred path is an asymmetry

The description explains this as preserved error checking for a bad `[CborConverter]` on a deferred member, but a member carrying that attribute returns early in `InitializeConverter`, before the new deferral check, so it can never reach the deferred branch. The verification in the getter therefore only ever applies to converters obtained from the registry — which the eager path (`_converter = _converterRegistry.Lookup(MemberType)`) does not verify at all.

The result is that a converter supplied through `RegisterConverter(Type, ICborConverter)` or a custom provider that does not implement `ICborConverter<MemberType>` now throws a `CborException` when it happens to sit on a cycle member, and an `InvalidCastException` otherwise. Please drop the verify from the deferred path so both paths behave the same.

---

## Non-blocking

### 3. Is `IsUnderConstruction` needed at all?

Once `MemberConverter` and `StructMemberConverter` resolve lazily, `MemberMapping` could skip the lookup during initialization entirely and always resolve in the getter. Nothing else in the library reads `IMemberMapping.Converter` — the only two consumers are the two member converter constructors this PR already makes lazy. That would remove the thread-static set, the `(registry, Type)` keying, the new `internal` registry API and the deferral flag, and it would cover every cycle shape rather than only those routed through a member mapping.

The cost is error timing: `No converter found for type X` would move from converter construction to first read/write. Since converter construction is itself lazy — it happens on the first serialization of the owning type — that window is narrow. Your version is the more conservative one and I am not asking you to change it, but if the simpler variant holds up under the full suite I would prefer it.

### 4. `CreateConverter`'s `finally` removes unconditionally

`underConstruction.Add((this, type))` returns `false` when the type is already in flight, but the `finally` removes the entry regardless. A re-entrant inner frame therefore clears the outer frame's in-flight marker while the outer constructor is still running. It is not harmful today — by the time the inner `finally` runs, the inner `GetOrAdd` has registered the converter, so later lookups hit the dictionary — but the invariant is fragile enough to be worth making explicit:

```csharp
bool added = underConstruction.Add((this, type));

try { ... }
finally
{
    if (added)
    {
        underConstruction.Remove((this, type));
    }
}
```

### 5. Eager lookups outside member mappings

`TupleConverter` (all arities) and `NullableConverter` still call `Lookup` from their constructors. Collections are already fine — `AbstractCollectionConverter`, `ArrayConverter` and `AbstractDictionaryConverter` use `field ??=`. For `class A { public Tuple<A> Pair; }` the recursion does terminate, because the second `CreateConverter(A)` is where the member defers, but it constructs and discards a duplicate `ObjectConverter<A>`, and the cached member mapping can be left pointing at the discarded instance. Same observable behaviour, wasted work. A comment noting the limit would be enough.

### 6. Tests

The four cases are the right ones, and the argument for why they could not ship alongside #152 holds — a `StackOverflowException` cannot be asserted on, so these only become writable once the fix exists. Three gaps worth filling:

* **A cycle through a collection**: `Parent { List<Child> }` / `Child { Parent }` is the shape most user models actually have, and it currently passes because of `AbstractCollectionConverter`'s laziness rather than because of anything in this PR. Pinning it keeps a future change there from silently reintroducing the overflow.
* **`StructMemberConverter`** is modified but no test exercises a cycle through it. Struct-to-struct cycles are impossible in C#, but `struct S { public Node N; }` / `class Node { public S S; }` is a genuine mutual reference and would cover the second half of the `MemberConverter.cs` change.
* The three-type ring is written but never read back.

---

No wire-format change, the same converters are selected for the same types, and the object-graph cycle staying a `MaxDepth`-bounded `CborException` is the right line to draw — thank you for pinning that as a test so the two failure modes do not get conflated later.

---

mcatanzariti (approved):
Approving `b509d54`. Taking §3 rather than patching §1 and §2 individually was the right call — both blockers disappear structurally instead of being made correct by argument, and the change is now 50 lines of source in two files.

What I verified rather than took on trust:

* `CborConverterRegistry.cs` is byte-identical to the base commit. The thread-static set, the `(registry, Type)` keying, the `internal IsUnderConstruction` API and the `try`/`finally` are all gone.
* **§1** — there is no flag left to publish out of order, and the single resolution site sits under the same `lock (this)` that `EnsureInitialize` uses. **§2** — one lookup site cannot be asymmetric with itself. Both correctly closed.
* The byte expectations in the three new hex-level tests (`Team`/`Member`, `Node`/`Slot`, `Boxed`) decode to what the comments say they do.
* This PR's own `build_and_test` is green now that the workflow run has been approved, alongside your fork run. 982 = 974 + 8 lines up with the 978 = 974 + 4 from the previous head.

**You are right about §5 and I was wrong.** `TupleConverterProvider` matches only the `ValueTuple<...>` generic definitions, and its `switch` has no `case 1`, so `class A { public Tuple<A> Pair; }` never reaches `TupleConverter` at all — it is mapped as an ordinary object with an `Item1` member. `ValueTuple<Boxed, int>` is the shape that actually exercises those eager lookups, and pinning it with the bytes is better coverage than what I described.

Three notes, none blocking, none of which need action in this PR:

1. `lock (this)` is now held across a `Lookup` that can construct an entire converter graph, and that construction takes further `MemberMapping` locks. Since I proposed this shape I went looking for an ABBA deadlock in it. It holds: reaching the getter requires the owning converter to be registered, and a registered type is never reconstructed, so the reverse edge — a thread holding a member mapping lock while constructing the owner — cannot exist. Worth a sentence of comment if you ever revisit the locking here, since the invariant is what makes it safe rather than the lock ordering.
2. The move of `No converter found for type X` to first read/write is real and, as you say, unobserved by the suite in either direction. Release-note material rather than something to fix.
3. Possible follow-up, out of scope here: the `parentConverter` reuse path from #147 has become redundant. With lazy resolution the direct `A.Self` case resolves through the getter, and doing so is marginally safer — the getter yields the registry's instance, whereas the reuse path can bind a member to an `ObjectConverter` that is still being constructed and may lose the `GetOrAdd` race.

Thanks for the thorough follow-up, and for flagging which of your claims were arguments from call order rather than measurements — that distinction made this much quicker to check.

---

PR #161, opened 2026-08-05
Stop the read path's probes from consuming semantic tags (split from #157)

Rafael-SOWNet:
This is the read-path half of #157, split out as you asked, with the B4 fix included. It stands on its own: no typed arrays, no new public API, no change to written bytes in any configuration.

## What it claims

Four probes in the read path call a *read* entry point to decide what comes next. Every read entry point begins with `SkipSemanticTag()`, so each of those probes consumes a tag that belongs to the value behind it. This stops that.

It does **not** change nested-tag behaviour, in either direction. Nested semantic tags are unsupported before and after; `NestedSemanticTagsAreNotSupported` records that rather than leaving it undiscovered.

Nothing in the library depended on the old behaviour, because the invariant that hides it is real: a tag left in front of a value is harmless, since each converter skips its own. It only becomes a bug once a converter must *see* its own tag — which is what #157 needs, and why this is separable from it.

| Probe | Was | Now |
|---|---|---|
| `MemberConverter.Read`, `StructMemberConverter.Read` — null check | `GetCurrentDataItemType()` | `IsNull()` |
| `MoveNextMapItem`, `ReadArray`, `SkipArray`, `SkipMap`, and 35 sites in `TupleConverter` — break check | `GetCurrentDataItemType()` | `IsBreak()` |
| `ObjectConverter`, `CborObjectFormat.Array` — discriminator tag test | `TryReadSemanticTag`, result discarded | non-discriminator tag returned to the reader |
| `TupleConverter` — a tuple's own tag | failed on it | skips it |

`IsNull()` and `IsBreak()` are `internal` header inspections that skip nothing. `IsNull()` matches `undefined` as well as `null`, because `GetCurrentDataItemType()` reports both as `CborDataItemType.Null` and the probe replacing it has to agree.

## B4 — measured

You asked for a number rather than an assurance, and it turned out to matter more than I expected.

`IsNull()` is deliberately not a bookmark-and-rewind. It runs once per member of every object read, and `CborReaderBookmark` carries a `ReadOnlySpan<byte>`, a `ReadOnlySequence<byte>?` and a `SequenceReader<byte>` — copying that twice per member is a cost every caller pays, including those who never meet a semantic tag.

net10.0 Release, 1M deserializations of a six-member class, minimum of seven measured passes after a 200k warmup (CoreCLR tier-0 code is unoptimised, so a short warmup measures the wrong thing):

| Member null probe | ns/object | ns/member | vs. `master` |
|---|--:|--:|--:|
| `master`, `GetCurrentDataItemType()` | 324–330 | ~54 | — |
| bookmark-and-rewind | 464–467 | ~77 | **+43%** |
| `IsNull()`, as written here | 315–316 | ~53 | **−3%** |

The bookmark costs ~43% of object deserialization. `IsNull()` removes it and lands marginally *under* `master`, because it also skips the `SkipSemanticTag()` call that `GetCurrentDataItemType()` performs.

The one remaining bookmark is `ObjectConverter`'s discriminator probe, which is once per object and only in `CborObjectFormat.Array` — not a per-member cost.

## What changes behaviour

**One value shape.** A null or `undefined` behind a semantic tag no longer registers with the member probe, so `RequirementPolicy.DisallowNull` does not reject that shape. The value is unaffected: the member's own converter calls `ReadNull()`, which skips the tag and yields null. Pinned by `ATaggedNullIsNotRejectedByDisallowNull` and `ATaggedUndefinedIsNotRejectedByDisallowNull`, with `AnUntaggedNullIsStillRejectedByDisallowNull` and `AnUndefinedMemberIsRejectedByDisallowNull` alongside them for the ordinary cases.

**Malformed input inside an indefinite-length container fails later, and differently.** `GetCurrentDataItemType()` throws `Primitive not supported` for primitives it has no mapping for, such as a bare `SimpleValue`; `IsBreak()` returns false and defers to the item's own converter. The document is still rejected, but by a different call and with a different message. Probably an improvement — the error now names what the converter expected — but it is a change, not a no-op.

**`MoveNextMapItem` is `public`.** Its tag-consumption behaviour changes for external callers who drive the reader themselves, not only for the library's own converters. Nothing else on the public surface moves.

## Tests

`SemanticTagProbeTests.cs`, 16 cases: class and struct members, definite- and indefinite-length arrays and maps, `Array` object format with a non-discriminator tag on the first item both with and without a discriminator registered, tagged tuples, tuple item tags, all four `DisallowNull` shapes (`null` and `undefined`, tagged and not), nested tags, and that writing is unaffected. 1088 → 1104 on net10.0; all four TFMs build.

Two of them fail without the source change and would not have without care:

* `ArrayFormatKeepsAFirstItemTagWhenADiscriminatorIsExpected` is the one that reaches the `ObjectConverter` bookmark. The whole `case CborObjectFormat.Array:` block runs only when `DiscriminatorConventionRegistry.GetConvention` returns a convention for the type being read, so `ArrayFormatKeepsAFirstItemTagWithoutADiscriminator` — which is the same shape on an undiscriminated type — passes with that hunk reverted, on the `MemberConverter` fix alone. Both are kept, and the names now say which path each covers.
* `AnUndefinedMemberIsRejectedByDisallowNull` fails if `IsNull()` matches only `Null`. That is a lost check rather than a changed message: `CborValueConverter` maps `F7` to `CborValue.Null`, a non-null reference, so `MemberConverter.Write`'s later null check never sees it either.

`NestedSemanticTagsAreNotSupported` now asserts the message, not just the exception type — otherwise any unrelated regression that throws keeps it green.

Also included, since it is on lines this touches: `Tuple6Converter`'s T5 break check said `Expected CBOR Array of size 5`, the one arity message in that file naming the wrong size. Pre-existing on `master`.

## CI

Green on this head, upstream, on .NET 8, 9 and 10 — 1104 tests on each:

https://github.com/dahomey-technologies/Dahomey.Cbor/actions/runs/31081457392

Rebased onto `master` at `563273f`, so that run is against the tree with #158 and #160 in it. No conflicts.

---

Once this lands I'll rebase #157 on top; what remains there is B1 and B5, plus the B2 README item.

### DIFF
--- a/src/Dahomey.Cbor.Tests/SemanticTagProbeTests.cs
+++ b/src/Dahomey.Cbor.Tests/SemanticTagProbeTests.cs
@@ -1,0 +1,293 @@
+﻿using System;
+using System.Collections.Generic;
+using Dahomey.Cbor.Attributes;
+using Dahomey.Cbor.Serialization;
+using Dahomey.Cbor.Serialization.Converters;
+using Dahomey.Cbor.Tests.Extensions;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests
+{
+    /// <summary>
+    /// A semantic tag in front of a value must survive the read path's probes and reach the
+    /// converter that owns it.
+    /// </summary>
+    /// <remarks>
+    /// Every <see cref="CborReader"/> read entry point begins with <c>SkipSemanticTag()</c>, so for
+    /// converters that ignore tags a probe consuming one is invisible: the value still decodes. It
+    /// stops being invisible as soon as a converter needs to <em>see</em> its tag, which is what
+    /// <see cref="TagObservingConverter"/> does here — it is the smallest thing that can tell the
+    /// difference, and it stands in for any converter decoded from its tag.
+    /// <para>
+    /// Four probes consumed a tag while only meaning to look at one: the null check in
+    /// <c>MemberConverter</c> and <c>StructMemberConverter</c>, the discriminator test in
+    /// <c>ObjectConverter</c>'s Array format, the break checks in <c>MoveNextMapItem</c>,
+    /// <c>ReadArray</c>, <c>SkipArray</c> and <c>SkipMap</c>, and the same break check across the
+    /// tuple converters.
+    /// </para>
+    /// </remarks>
+    public class SemanticTagProbeTests
+    {
+        [CborConverter(typeof(TagObservingConverter))]
+        public class Tagged
+        {
+            public ulong Tag { get; set; }
+            public int Value { get; set; }
+        }
+
+        /// <summary>
+        /// Reads its own semantic tag instead of letting the reader skip it, so a probe that
+        /// consumed the tag shows up as <c>Tag == 0</c> rather than as a decoding failure.
+        /// </summary>
+        public class TagObservingConverter : CborConverterBase<Tagged>
+        {
+            public override Tagged Read(ref CborReader reader)
+            {
+                ulong tag = reader.TryReadSemanticTag(out ulong semanticTag) ? semanticTag : 0;
+
+                return new Tagged { Tag = tag, Value = reader.ReadInt32() };
+            }
+
+            public override void Write(ref CborWriter writer, Tagged value, LengthMode lengthMode)
+            {
+                writer.WriteSemanticTag(value.Tag);
+                writer.WriteInt32(value.Value);
+            }
+        }
+
+        public class Holder
+        {
+            public Tagged Member { get; set; }
+        }
+
+        public struct StructHolder
+        {
+            public Tagged Member { get; set; }
+        }
+
+        /// <summary>
+        /// A string member, so the null case is about the probe rather than about whether the
+        /// member's converter happens to handle null.
+        /// </summary>
+        public class RequiredHolder
+        {
+            [CborRequired(RequirementPolicy.DisallowNull)]
+            public string Member { get; set; }
+        }
+
+        /// <summary>
+        /// The Array object format keys members by position, so every member needs an explicit index.
+        /// </summary>
+        public class ArrayHolder
+        {
+            [CborProperty(0)]
+            public Tagged Member { get; set; }
+        }
+
+        public class IntHolder
+        {
+            public int Value { get; set; }
+        }
+
+        /// <summary>
+        /// The null probe on a class member. This is every object member in the library, so before
+        /// the fix no tag-decoded value could be read back as a member of anything.
+        /// </summary>
+        [Fact]
+        public void AClassMemberKeepsItsSemanticTag()
+        {
+            // a1                     map(1)
+            //    66 4d656d626572     "Member"
+            //    c1                  tag(1)
+            //    05                  5
+            Holder holder = Helper.Read<Holder>("A1664D656D626572C105");
+
+            Assert.Equal(1UL, holder.Member.Tag);
+            Assert.Equal(5, holder.Member.Value);
+        }
+
+        /// <summary>Same probe, struct member path.</summary>
+        [Fact]
+        public void AStructMemberKeepsItsSemanticTag()
+        {
+            StructHolder holder = Helper.Read<StructHolder>("A1664D656D626572C105");
+
+            Assert.Equal(1UL, holder.Member.Tag);
+            Assert.Equal(5, holder.Member.Value);
+        }
+
+        /// <summary>
+        /// The break probe in <c>ReadArray</c>. A break marker is never tagged, so a tag here belongs
+        /// to the next item and must survive.
+        /// </summary>
+        [Fact]
+        public void AnIndefiniteLengthArrayKeepsItemTags()
+        {
+            // 9f            array(*)
+            //    c1 05      tag(1) 5
+            //    c2 06      tag(2) 6
+            // ff            break
+            List<Tagged> items = Cbor.Deserialize<List<Tagged>>("9FC105C206FF".HexToBytes());
+
+            Assert.Equal(2, items.Count);
+            Assert.Equal(1UL, items[0].Tag);
+            Assert.Equal(5, items[0].Value);
+            Assert.Equal(2UL, items[1].Tag);
+            Assert.Equal(6, items[1].Value);
+        }
+
+        /// <summary>The break probe in <c>MoveNextMapItem</c>, on an indefinite-length map.</summary>
+        [Fact]
+        public void AnIndefiniteLengthMapKeepsValueTags()
+        {
+            // bf                     map(*)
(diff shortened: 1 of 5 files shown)


mcatanzariti (commented):
Reviewed statically against `6f64329` (no .NET SDK available in my environment, so I did not run the suite). Note that `build_and_test` has since run on this PR and is green, so the CI caveat in the description is resolved.

## What I verified

I checked the mechanism rather than taking the description on trust, and it holds up:

- **`IsNull()`/`IsBreak()` really are non-destructive.** `GetHeader()` (`CborReader.cs:914`) consumes the header byte into `_header` and sets `_state = Header`; the next `GetHeader()` returns the cache, and `Advance()` flips state back to `Data`. That is exactly what the pre-existing `Accept()` helper at `CborReader.cs:1100` does, so this reuses an established idiom rather than introducing one.
- **"`TupleConverter` is the only converter that reaches the reader below its tag-skipping entry points"** — confirmed. It holds the only seven `ReadSize()` call sites outside `CborReader`; `ReadArray`/`ReadMap` open with `ReadBeginArray`/`ReadBeginMap`, which skip tags. So the added `TryReadSemanticTag(out _)` is genuinely required to avoid the regression described, and the 35 `IsBreak` sites = 2+3+…+8 checks out.
- **The `ObjectConverter` bookmark is safe.** Only `TryReadSemanticTag` runs between `GetBookmark()` and `ReturnToBookmark()`, and it does not touch `_depth` — which matters, because `CborReaderBookmark` does *not* capture `_depth` (a pre-existing gap since #155; harmless here, but a trap for future bookmark users).
- The `context.memberIndex++` relocation into the new scope block is behaviour-preserving: it ran after the if/else before, and still does.

## Two things I would like addressed before merge

**1. There is a second undocumented behaviour change, not one.**

`GetCurrentDataItemType()` maps *both* `CborPrimitive.Null` and `CborPrimitive.Undefined` to `CborDataItemType.Null` (`CborReader.cs:207-209`). `IsNull()` matches only `Null`. So an `undefined` (0xf7) member value no longer trips `RequirementPolicy.DisallowNull`/`Always` either.

For a `string` member the failure just changes shape — "Expected major type TextString" instead of "Property 'X' cannot be null" — but for a member whose converter tolerates `undefined`, the check is simply gone.

Unlike the tagged-null case, there is no argument in favour of this one, and it is free to fix:

```csharp
return header.MajorType == CborMajorType.Primitive
    && (header.Primitive == CborPrimitive.Null || header.Primitive == CborPrimitive.Undefined);
```

That costs nothing on the hot path and restores parity with the old probe. No test currently covers it.

**2. The `ObjectConverter` change is not exercised by its test.**

`ArrayFormatKeepsATagOnTheFirstItemWhenItIsNotTheDiscriminator` uses `ArrayHolder`, which carries no `[CborDiscriminator]`. `DiscriminatorConventionRegistry.InternalGetConvention` returns null when `objectMapping.Discriminator == null`, so `_discriminatorConvention` is null and the whole `case CborObjectFormat.Array:` block in `ReadItem` (`ObjectConverter.cs:519`) is never entered.

The test passes purely because of the `MemberConverter` fix — it would still pass with the `ObjectConverter` hunk reverted. Exercising the bookmark path needs a discriminated hierarchy in `Array` object format (the setup at `IntDiscriminatorTests.cs:502-553`) with a non-discriminator tag on the first item.

## Worth stating in the description

- **The break probes loosen error behaviour.** `GetCurrentDataItemType()` throws `"Primitive not supported"` for primitives such as `SimpleValue`; `IsBreak()` returns false and defers to the item's converter. Malformed input inside an indefinite-length container now fails later and with a different message. Probably an improvement, but it is a change.
- **`MoveNextMapItem` is public API.** Its tag-consumption behaviour changes for external callers, not only for the library's own converters.

## Nits

- `ATaggedNullIsNotRejectedByDisallowNull` has its 4-line hex annotation pasted twice.
- `NestedSemanticTagsAreNotSupported` asserts only `CborException`, so any unrelated regression that throws keeps it green. Pinning the bug as documentation is defensible; asserting on the message would actually pin the over-skip.
- While touching those exact lines: `Tuple6Converter` still says `"Expected CBOR Array of size 5"` for its T5 break check (pre-existing on `master`). Free drive-by.

## Overall

Mechanically sound and well-scoped: `internal` helpers only, no public API added, write path untouched, one commit. The reasoning in the description matches the code. The rest is documentation.

---

mcatanzariti (approved):
All three items are addressed. Re-reviewed statically at `c74ce43` (still no .NET SDK on my side, so I did not run the suite; `build_and_test` is green on this head).

**`IsNull()` / `undefined`** — the justification is sharper than my original note. I wrote that the check "disappears for a member whose converter tolerates `undefined`"; you identified the concrete shape: `CborValueConverter.ReadCborValue` maps `CborDataItemType.Null` — which covers `F7` — to `CborValue.Null`, a non-null reference, so nothing downstream re-raises the requirement. That makes it a lost check rather than a reworded failure, and `AnUndefinedMemberIsRejectedByDisallowNull` fails without the one-line change. The two companion tests pin the other directions.

**`ArrayFormatKeepsAFirstItemTagWhenADiscriminatorIsExpected`** — this one binds. I traced it: the `[CborDiscriminator]` makes `GetConvention` return a convention, so the `case CborObjectFormat.Array:` block is entered; `TryReadSemanticTag` finds tag 1 ≠ 39, `ReturnToBookmark` hands it back, and the member at index 1 reads it. With the `ObjectConverter` hunk reverted the tag is discarded and `Tag` is 0, so the assertion fails. The `[CborProperty]` indexes starting at 1 are right, given index 0 is the discriminator's slot. Renaming the older test to say which path it covers is a good call.

**`NestedSemanticTagsAreNotSupported`** — the asserted message checks out: in `A1 65 "Value" C1 C0 01`, the second tag is met where a value is expected at offset 9, so `"[9] Invalid major type SemanticTag"` is exactly right. The rewritten remark is also a correction I had missed — the old one attributed the failure to `GetCurrentDataItemType`'s recursion, which is not the path taken here; a read entry point performing a single `SkipSemanticTag()` is.

Plus the `Tuple6Converter` arity message and the duplicated hex annotation.

---

One thing I noticed while checking the `CborValue` reasoning, **pre-existing on `master` and out of scope here** — filing it separately rather than holding this up. `ReadNull()` is `SkipSemanticTag(); return Accept(CborPrimitive.Null);`, which returns `false` for `F7` and leaves `_state = Header` with the `F7` header cached and the position already past the byte. `ReadCborValue` returns `CborValue.Null` anyway, so the item is never consumed and the reader re-presents that header to the next read. It is invisible when the `undefined` is the last item of its container — which is the case in `AnUndefinedMemberReadsWhenThePolicyAllowsIt` — but not otherwise.

LGTM.

---

PR #165, opened 2026-08-06
Emit semantic tags when writing the object model, closes #162

Rafael-SOWNet:
Fixes #162. Two commits, because fixing what the issue describes turned up a defect underneath it that had to be fixed first.

## The one underneath — please read this part first

**Reading a tagged value mutates an instance shared by the whole process.** `CborValueConverter.Read` assigned the captured tag straight onto whatever `ReadCborValue` returned, and several value types hand out shared instances: `CborPositive` and `CborNegative` for integers 0–100, `CborSingle`/`CborDouble`/`CborDecimal` for small whole numbers, `CborBoolean` for both of its values, and `CborValue.Null`, which is one instance for the process.

Run against `master`:

```
Cbor.Deserialize<CborValue>("C101".HexToBytes());          // tag(1) 1, in some document

Cbor.Deserialize<CborArray>("820102".HexToBytes())         // another document, no tags at all
new CborArray { 1, 2 }                                     // built in code, never read
CborValue.Null.SemanticTag                                 // whatever a tagged null last set
```

the second and third both come back with `SemanticTag == 1` on their first item, and `CborValue.Null.SemanticTag` is `2` after anything anywhere reads a tag-2 null.

This is live on `master` today — anyone inspecting `SemanticTag` already gets another document's answer. It is invisible only because nothing reads it back. **Emitting tags on write is what turns it into wrong bytes**, and it would have done so for the most common values there are: with only the second commit, that unrelated `820102` document serializes as `82 C101 02`.

`649b4db` fixes it with an internal `CborValue.WithSemanticTag`, virtual, tagging in place for the types constructed fresh per value and overridden by the seven with shared instances to copy first. No public API change. Three tests, each failing without it.

If you would rather have that as its own PR against `master` I will split it out — it stands alone, and #162's fix should not merge without it.

## The issue itself

`fb8a562`. `Read` captured the tag, `Write` switched on `Type` and never read it back, so a DOM read-then-write dropped every tag: tag 0/1 a datetime became a bare string or integer, 2/3 a bignum became a byte string, 32 a URI became text, 39 a polymorphic document lost its type, 64–87 a typed array became opaque bytes.

**The read side had a matching gap**, which the issue does not mention: reaching the container converters directly — `Cbor.Deserialize<CborObject>(…)`, which is what the README shows — bypasses `Read(CborValue)` entirely, and their own `Read` never captured a tag. So a tagged map or array lost it on the way *in* as well as out, and the round trip needed both halves.

On your two open decisions:

1. **Where the tag is written** — by whoever writes the value, exactly once. Containers emit their own rather than having it emitted for them, because those writers are reachable without passing through the `CborValue` switch; doing both would double the tag on that route. Nested values are asserted rather than assumed — `ATagNestedTwoLevelsDeepSurvives` puts a bignum inside an array inside a map.
2. **Round-trip fidelity vs. correctness on edit** — re-emitted unconditionally, which is what makes the round trip lossless. A caller who reads a tag-1 value and replaces it with a string does get a tag that no longer matches, but an edited value is indistinguishable from an untouched one, so dropping tags on suspicion would break the case this exists for. The reasoning is on the method rather than left to be re-derived.

I did **not** touch `SemanticTag`'s `internal set`. Opening it is a public API decision and it is separable from the round-trip bug now that the round trip works.

## Tests

Twelve for the write fix, three for the sharing fix. Every tagged shape the issue lists is a theory case, and `AnUntaggedDocumentIsUnchanged` pins byte-identity for callers who never meet a tag.

1121 tests pass on net10.0, up from 1104. All four TFMs build.

## Note for #157

That PR pins the current behaviour with `TheObjectModelSeesATypedArrayAsAByteStringAndDropsItsTag`. Whichever of these merges second needs that test updated — it is a one-line expectation change, and I will do it on whichever side is still open.

### DIFF
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0162.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0162.cs
@@ -5,23 +5,170 @@ using Xunit;
 namespace Dahomey.Cbor.Tests.Issues
 {
     /// <summary>
-    /// Issue #162, first half: a semantic tag captured on read must not attach itself to every other
-    /// occurrence of that value in the process.
+    /// Issue #162: reading a document into the object model and writing it back dropped every semantic
+    /// tag it carried.
     /// </summary>
     /// <remarks>
-    /// <see cref="CborPositive"/> and <see cref="CborNegative"/> hand out shared instances for small
-    /// integers, as <see cref="CborSingle"/>, <see cref="CborDouble"/> and <see cref="CborDecimal"/> do
-    /// for small whole numbers, <see cref="CborBoolean"/> for both values and
-    /// <see cref="CborValue.Null"/> for null. Assigning <c>SemanticTag</c> to one of those attaches the
-    /// tag to that value everywhere — in documents that never carried a tag, and in values built in
-    /// code that were never read at all.
+    /// <c>CborValueConverter.Read</c> captures the tag into <see cref="CborValue.SemanticTag"/>, but
+    /// the write path switched on <see cref="CborValue.Type"/> and never read it back, so the tag was
+    /// silently lost. The tag is often what carries the meaning — tag 1 makes an integer an epoch
+    /// datetime, tag 2 makes a byte string a bignum, tag 39 carries a discriminator — so dropping it
+    /// changed what the document said rather than merely losing an annotation, and raised no error.
     /// <para>
-    /// This is observable today through the property, and it is what would turn emitting tags on write
-    /// into wrong bytes rather than a fix.
+    /// Tags are re-emitted unconditionally, which is what makes a read-then-write lossless. A caller
+    /// that reads a tag-1 value and replaces it with a string does get a document whose tag no longer
+    /// matches its content, but that is the caller describing their own value: the alternative,
+    /// dropping tags whose content changed, cannot be distinguished from the round trip this fixes.
     /// </para>
     /// </remarks>
     public class Issue0162
     {
+        /// <summary>The repro from the issue: a tagged epoch datetime.</summary>
+        [Fact]
+        public void ATaggedValueSurvivesARoundTrip()
+        {
+            // c1                  tag(1)
+            //    1a 514b67b0      1363896240
+            const string hexBuffer = "C11A514B67B0";
+
+            CborValue value = Cbor.Deserialize<CborValue>(hexBuffer.HexToBytes());
+
+            Assert.Equal(1UL, value.SemanticTag);
+            Helper.TestWrite(value, hexBuffer);
+        }
+
+        /// <summary>
+        /// A tag on a value nested inside a map, which reaches the write path through
+        /// <c>WriteMapItem</c> rather than the top-level call.
+        /// </summary>
+        [Fact]
+        public void ATagOnAMapValueSurvives()
+        {
+            // a1                  map(1)
+            //    61 61            "a"
+            //    c1 1a 514b67b0   tag(1) 1363896240
+            const string hexBuffer = "A16161C11A514B67B0";
+
+            CborObject obj = Cbor.Deserialize<CborObject>(hexBuffer.HexToBytes());
+
+            Assert.Equal(1UL, obj["a"].SemanticTag);
+            Helper.TestWrite(obj, hexBuffer);
+        }
+
+        /// <summary>A tag on a map <em>key</em>, which is written by the same call as the value.</summary>
+        [Fact]
+        public void ATagOnAMapKeySurvives()
+        {
+            // a1                  map(1)
+            //    c1 01            tag(1) 1
+            //    02               2
+            const string hexBuffer = "A1C10102";
+
+            CborObject obj = Cbor.Deserialize<CborObject>(hexBuffer.HexToBytes());
+
+            Helper.TestWrite(obj, hexBuffer);
+        }
+
+        /// <summary>A tag on an array item, reached through <c>WriteArrayItem</c>.</summary>
+        [Fact]
+        public void ATagOnAnArrayItemSurvives()
+        {
+            // 82                  array(2)
+            //    c1 01            tag(1) 1
+            //    02               2
+            const string hexBuffer = "82C10102";
+
+            CborArray array = Cbor.Deserialize<CborArray>(hexBuffer.HexToBytes());
+
+            Assert.Equal(1UL, array[0].SemanticTag);
+            Helper.TestWrite(array, hexBuffer);
+        }
+
+        /// <summary>
+        /// A tag on the container itself. The map and array writers are reachable directly — a member
+        /// declared <see cref="CborObject"/> does not go through the <see cref="CborValue"/> switch —
+        /// so they have to emit the tag themselves rather than rely on their caller.
+        /// </summary>
+        [Fact]
+        public void ATagOnAMapItselfSurvives()
+        {
+            // c1                  tag(1)
+            //    a1 6161 01       {"a": 1}
+            const string hexBuffer = "C1A1616101";
+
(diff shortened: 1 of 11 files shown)


mcatanzariti (commented):
## Analysis

The fix is correct and the design calls are the right ones. One thing blocks a clean merge, and it isn't what CI says.

Up front: I could not build or run the suite in this environment — no .NET SDK available and the sandbox network policy blocks the SDK download. Everything below is static analysis against the merged tree plus a trial merge.

### The blocker: green CI is stale

This branch is based on `3dd8989`. `master` has since moved to `1190e6d` (#157, typed arrays). Both check runs passed — against the old base.

GitHub reports `mergeable_state: clean`, and a trial merge into current `master` produces **zero textual conflicts**. But there is a semantic conflict, the one flagged in the PR description:

`src/Dahomey.Cbor.Tests/TypedArrayTests.cs:776` — `TheObjectModelSeesATypedArrayAsAByteStringAndDropsItsTag`, which arrived with #157 and pins the old behaviour:

```csharp
Assert.Equal(CborValueType.ByteString, value.Type);   // still passes
Assert.Equal(85ul, value.SemanticTag);                // still passes
Helper.TestWrite(value, "480000C03F00002040", null, options);   // now fails
```

After this PR that write emits `D855480000C03F00002040`. One expectation line plus the comment above it. #157 did not touch `CborValueConverter.cs` at all, so that test is the only interaction between the two changes.

### The underlying bug is real, and leading with it was right

Verified in source rather than taken on trust:

* `CborPositive` / `CborNegative` — `_preCreatedValues`, integers 0–100
* `CborSingle` / `CborDouble` / `CborDecimal` — same pattern for small whole numbers
* `CborBoolean` — static `_true` / `_false`
* `CborValue.Null` — one `CborNull` for the process

`Read` assigned `cborValue.SemanticTag = semanticTag` onto whatever `ReadCborValue` returned. That is process-wide mutation, live on `master` today. Emitting tags on write without fixing it first would have turned it into wrong bytes for the most common values in any document.

The `WithSemanticTag` fix is **exactly complete**: 11 `CborValue` subclasses exist, 7 cache instances, and those same 7 override. `CborString`, `CborByteString`, `CborObject` and `CborArray` construct fresh per value and correctly take the in-place base. `SemanticTag`'s setter is `internal`, and the only assignments left in the library are the 7 overrides, the base method, and the two container reads — both on freshly constructed objects. No leak path remains.

One risk the copy introduces, checked: `CborNull.WithSemanticTag` returns a new instance, breaking reference identity with `CborValue.Null`. Nothing in the library or tests compares against it by reference, and `CborNull.Equals` is type-based. Safe.

### The write path holds up

The no-double-tag reasoning is subtle but sound. The load-bearing detail is that the tag went into the **3-arg** explicit interface implementations only — the 2-arg overloads delegate to the `CborValue` switch, which skips `Object`/`Array`, which casts back into the 3-arg implementation. One tag on every route, including the direct-container route that motivated the split.

Two things that could have broken and don't:

* **Deterministic encoding** (#158) — `DeterministicKeyOrder` sorts keys by encoding them through the same converter that writes them, so tags on keys now appear in both the sort key and the output, consistently. That is RFC 8949 §4.2.1 read correctly: the tag is part of the encoded data item.
* **Ordering of `TryReadSemanticTag` before `ReadNull`** — necessary, not stylistic. `ReadNull()` calls `SkipSemanticTag()`, so the other order consumes and discards the tag. This only works because #161 stopped the member null probe eating tags first.

All 14 hex vectors in the tests hand-verified, including that re-emitted tags use canonical head encoding (`C0`–`C3` for 0–3, `D820`/`D827`/`D855` for 32/39/85).

### Two things worth raising

**1. "Lossless" is slightly overstated — nested tags still drop.** `SkipSemanticTag` skips exactly one tag with no loop, and `TryReadSemanticTag` captures one. For `C1C201` the inner tag is consumed by `GetCurrentDataItemType` and lost; the round trip yields `C101`. An improvement over `master`'s `01` and not a regression, but `SemanticTag` being a single `ulong?` means the object model structurally cannot hold tag chains. Worth stating as a known limit rather than leaving "lossless" unqualified.

**2. A sibling of this bug is still live.** `SetValue<T>` is public on all six caching types:

```csharp
CborValue v = (CborPositive)5;      // shared instance
((CborPositive)v).SetValue(999);    // every 5 in the process is now 999
```

Same aliasing family, out of scope here, but it probably deserves its own issue.

### On the two open decisions

Both calls look right. Emitting from whoever writes the value, exactly once, is the only option that survives the direct-container route. And re-emitting unconditionally is correct because an edited value is genuinely indistinguishable from an untouched one — dropping tags on suspicion would break the case the change exists for.

The offer to split `649b4db` into its own PR is worth taking if the sequencing should be clean: it touches only `ObjectModel/` plus 11 lines of `Read`, carries 3 self-contained tests, and stands alone as a `master` bug fix.

### Suggested next step

Rebase onto current `master` and update the one `TypedArrayTests.cs` expectation, so CI runs against the real base.

---

mcatanzariti (commented):
Both points addressed correctly. Re-verified against the merged tree; nothing blocking left from my side.

Same caveat as before: still no .NET SDK in this environment and the network policy blocks the download, so this is static analysis plus trial merges, not a test run.

### The two fixes

**The pinned expectation.** `TypedArrayTests.cs` now asserts `Helper.TestWrite(value, hexBuffer, null, options)` with `hexBuffer = "D855480000C03F00002040"`, which is exactly what the new write path emits. The `ByteString` / `85ul` assertions still hold. The rename to `TheObjectModelSeesATypedArrayAsATaggedByteString` is right — it no longer pins a drop, and the comment rewrite says what it now pins.

**The nested-tag qualification.** Traced `C1C201` rather than taking the test's word for it: `TryReadSemanticTag` consumes `C1`, `GetCurrentDataItemType`'s `SkipSemanticTag` absorbs `C2`, the value comes back tagged 1 and re-writes as `C101`. `ANestedTagKeepsOnlyTheOuterOne` is correct, and putting the limit on the method rather than in a thread is the right call.

**The rebase preserved the first two commits.** The delta between the old and new heads across `ObjectModel/` and `CborValueConverter.cs` is exactly #164's change (`CborValueType.cs` +8, `CborValueConverter.cs` +9) — the base moving, not the content. The 7 `WithSemanticTag` overrides are intact, and #164's `undefined` fix composes cleanly with tag capture: `ReadCborValue` returns the shared `CborValue.Null` for `F7`, and `WithSemanticTag` copies before tagging, so the shared-null guarantee survives the new neighbour.

### `master` moved again — no action needed

This is based on `cea8461`; `master` is now `8ed2cc1` (#156, the AOT source generator). Verified rather than assumed:

* #156 touches neither `CborValueConverter.cs` nor `ObjectModel/`;
* the generator's only `SemanticTag` mention is `DiscriminatorSemanticTag` — the tag-39 discriminator option copied into generated `CborOptions`, unrelated to `CborValue.SemanticTag`;
* trial merge onto `8ed2cc1` is textually clean, and there is no second `TypedArrayTests`-style expectation hiding in it.

So no rebase this time.

### On splitting `649b4db` / `992853b`

Not needed — keep both commits here. The reasoning for the split still stands on its merits, but the sequencing problem it was meant to solve has gone away now that this is rebased and green, and two well-separated commits in one PR carry the history just as well as two PRs.

### `C1C2C301`

Filing it separately is right, and nothing open covers it yet. One caveat on the diagnosis before anyone acts on it: I could not reproduce the exception statically. Reading the path, `Advance(1)` looks correct for a tag whose argument is in the header byte, so if the failure is real the cause is more likely in the `_state` / `Accept` interaction than in the double-advance itself. Your run is better evidence than my reading, but the *cause* is worth re-establishing before a fix is written against it.

### #168

Good issue — the boolean case in particular makes the severity concrete in a way the integer one does not. Reply there rather than here.

---

PR #166, opened 2026-08-06
Read RFC 8949 indefinite-length byte and text strings

Rafael-SOWNet:
Reads RFC 8949 §3.2.3 indefinite-length byte and text strings, which the reader refused.

```csharp
Cbor.Deserialize<string>("7F 62'st' 64'ream' FF")   // was: NotSupportedException
                                                     // now: "stream"
```

This is core format rather than an extension — any encoder may emit it, and it is what a producer that does not yet know a value's length emits. `ReadSizeAndBytes`, `ReadSizeAndSequence` and `SkipSizeAndBytes` all called `ThrowNotSupported()` on a size of `-1`, which surfaced as a bare `NotSupportedException` that a caller's `catch (CborException)` does not catch.

## The library already disagreed with itself

`CborDataItemScanner` walks this shape correctly — it has since #154. So the scanner reported a document **`Complete`** that `Cbor.Deserialize` then threw on. `TheScannerAndTheReaderAgreeOnAChunkedString` pins that they no longer do, and it matters for #134: anything built on the scanner, which is the streaming primitive #139 asked for, meets this immediately.

## Decisions worth stating

**The chunks are joined before decoding, not decoded per chunk.** A multi-byte character may straddle a boundary — `7F 61E2 6282AC FF` is one EURO SIGN, and two invalid fragments to anything that decodes each chunk on its own. `AUtf8CharacterMaySpanAChunkBoundary` is that case.

**The result is a fresh array, not a slice.** The pieces are contiguous in neither the input nor the scratch buffer, and the scratch buffer is a single shared one the next chunk would overwrite. The allocation is confined to documents that use the encoding — a definite-length string does not reach the code, and `ReadSizeAndBytes` still returns a slice for it.

**Two shapes are malformed and refused rather than interpreted**, both as `CborException`: a chunk whose major type differs from the enclosing string, which would otherwise turn bytes into text nobody wrote, and a nested indefinite-length string. **Skipping validates them too** — a chunked value in a member nobody asked for must not pass unnoticed and derail the rest of the map, which `AChunkedStringIsSkippedCorrectly` pins.

**Read only.** Writing indefinite-length strings is a separate question — `LengthMode.IndefiniteLength` applies to arrays and maps, and nothing here changes what is written.

## Relationship to #157

#157's first commit makes these three sites throw a `CborException` instead of a `NotSupportedException`. This supersedes it: the sites no longer throw at all. If this lands first I will drop that commit from #157; if #157 lands first this rebases over it cleanly, since it replaces the same three bodies.

## Tests

`ChunkedStringTests.cs`, 17 cases: both major types, one, three and zero chunks, a character spanning a boundary, into a member and into the object model, skipped, as a `ReadOnlySequence`, inside an indefinite-length array so the string's break is not taken for the array's, both malformed shapes, and the scanner/reader agreement. `Helper.Read` runs the text cases through the span, sequence and fragmented readers, so the segmented path is covered rather than assumed.

1121 tests pass on net10.0, up from 1104. All four TFMs build.

## What this does not close

Two further gaps I confirmed by running them against `master`, neither touched here:

* **Duplicate map keys** (§5.6) throw `ArgumentException` from the backing dictionary — again outside the `CborException` contract.
* Tags 2–5 decode structurally (a bignum arrives as a byte string, a decimal fraction as an array) rather than semantically. That is permitted, but it is where `System.Formats.Cbor` is ahead.

Happy to take either if you want them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ High-performance [CBOR](https://cbor.io/) serialization framework for .Net (C#)
 * Support for RFC 8746 typed arrays, opt-in per direction (CborOptions.TypedArrayMode)
 * Support for RFC 8949 §3.4.3 bignums (tags 2 and 3) as System.Numerics.BigInteger
 * Duplicate map keys rejected on every decode target, with a last-wins opt-out (CborOptions.DuplicateKeyMode)
+* Reads RFC 8949 indefinite-length (chunked) byte and text strings; writes definite-length only
 
 ## Installation
 ### NuGet
@@ -206,11 +207,28 @@ definite-length byte string, so there is no array header to make indefinite. It 
 array not written as one.
 
 The object model does not model typed arrays. A `CborValue` holding one is a `ByteString` with the tag in
-`CborValue.SemanticTag`, so DOM code sees opaque bytes rather than numbers. Note that
-`CborValueConverter` does not write `SemanticTag` back — reading a document into a `CborValue` and
-serializing it again drops the tag, turning the typed array into a plain byte string. That is true of
-every semantic tag, not just these, but typed arrays are the easiest way to meet it. Tracked as
-https://github.com/dahomey-technologies/Dahomey.Cbor/issues/162.
+`CborValue.SemanticTag`, so DOM code sees opaque bytes rather than numbers. The tag itself survives a
+read-then-write, so the document still describes a typed array afterwards.
+
+### Indefinite-length strings (RFC 8949 §3.2.3)
+
+A byte or text string may arrive as a series of chunks terminated by a break, which is what a producer
+emits when it does not know the length in advance:
+
+```
+7F 62 7374 64 7265616D FF     reads as "stream"
+5F 42 0102 43 030405 FF       reads as the five bytes 01 02 03 04 05
+```
+
+An indefinite-length string denotes exactly the concatenation of its chunks, so it is indistinguishable
+from the definite-length string of the same content once read — there is nothing to configure and
+nothing to opt into.
+
+**Reading only.** Nothing is ever written in this form: `LengthMode.IndefiniteLength` applies to arrays
+and maps, not to strings, so what this library emits is unchanged.
+
+A chunk whose major type differs from the enclosing string, and a nested indefinite-length string, are
+both malformed and raise a `CborException`.
 
 ### Bignums (RFC 8949 §3.4.3)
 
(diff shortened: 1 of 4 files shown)


Rafael-SOWNet:
Both taken at `ff716f7`, rebased onto `1fdc7fc`. **1471 library tests plus 35 generator tests**, green on all three TFMs: https://github.com/dahomey-technologies/Dahomey.Cbor/actions/runs/31446702907

## 1 — the `Trim` change

You are right, and the direction of the mistake is the part worth naming: I optimised the path the review had just been discussing rather than the path that runs. `ReadByteStringSequence` on a chunked value is rare; `ReadString`, `ReadRawString` and `ReadByteString` are the read path.

Done as you described. `ReadChunksArray` hands back the buffer and the length, `Trim` is gone, and both callers take a window:

* `ReadChunks` → `new ReadOnlySpan<byte>(joined, 0, length)`
* `ReadSizeAndSequence` → `new ReadOnlySequence<byte>(joined, 0, length)`

Neither copies, on either path, which is better than what was there before the nit as well as after it. The `ByteBufferWriter` remark stayed with the allocation, and the doubling now carries a line saying why the buffer is handed over long rather than trimmed, so the next person to see the overshoot does not fix it back.

## 2 — the sequence path

Taken. `AChunkedByteStringReadsAsASequence` is now a `[Theory]` running each case through `new CborReader(buffer)` **and** `new CborReader(Helper.Fragmentize(buffer))` — and thank you for the note that `Helper.Read<ReadOnlySequence<byte>>` cannot reach it, which saved me finding out that the branch is unreachable through the registry.

**Your instinct that this should be pinned before the code changed was right, and for a sharper reason than "lower stakes now".** The case that was there could not have caught the change. `5F 42 0102 43 030405 FF` is two chunks of two and three bytes: the accumulator resizes to 2, then to `Math.Max(5, 4)` = 5, and lands exactly full — so `length == joined.Length` and handing the array back whole gives the same answer as windowing it. The test would have passed with the length ignored.

So the theory gains `5F 41 01 41 02 41 03 FF`, three one-byte chunks: 1 → 2 → `Math.Max(3, 4)` = 4, three bytes in a four-byte array. Confirmed it binds by making `ReadSizeAndSequence` return `new ReadOnlySequence<byte>(joined)` — one failure, that case, trailing zero. The zero-chunk case is in the theory too, since `Array.Empty` with a zero window is its own shape.

## The README hunk and #180

The rebase conflicted in exactly one place, the `## Features` list, where this PR's bullet meets the two `master` gained from #174 and #181; all three are kept. The semantic-tag paragraph merged clean, so the correction we agreed to keep here is what closes **#180** — I did not open a separate PR for it, since a second edit to the same five lines would conflict with this one.

## Where the branch stands

Base is `1fdc7fc`, `mergeable` is back to green, three commits. Nothing outstanding from either review round that I know of.

---

PR #167, opened 2026-08-06
Reject a duplicate map key as a CborException, not an ArgumentException

Rafael-SOWNet:
A CBOR map carrying the same key twice was already refused — but as an `ArgumentException`:

```
An item with the same key has already been added. Key: a
```

which a caller's `catch (CborException)` does not catch, and which names neither the key's position nor which document it came from. For anything decoding untrusted frames that is the difference between a rejected frame and an unhandled exception.

```
[7] Duplicate map key: "a"      // now
[3] A map key cannot be null.   // not a duplicate, and no longer reported as one
```

The position is the reader's position at the throw, i.e. the end of the entry rather than the start of the key — consistent with every other exception in the reader.

## Scope

**Three** read paths reach a `Dictionary.Add` while decoding, and all three are fixed:

| Path | Shape |
|---|---|
| `CborValueConverter` | the object model |
| `AbstractDictionaryConverter` | every typed dictionary — `Dictionary`, `IDictionary`, `SortedDictionary`, `ImmutableDictionary` |
| `ObjectConverter` | **types with a creator mapping** — records and `[CborConstructor]`, six call sites across both object formats |

The third was missed in the first round, and the survey that supported it was too narrow: it exercised a POCO with settable properties, which is a different branch entirely. A record was still failing with a bare `ArgumentException`.

Two further failures share the catch and must not be reported as duplicates: a **null key**, since `ArgumentNullException` derives from `ArgumentException`; and a **caller-supplied `IDictionary` refusing for its own reasons**, which is distinguished by probing `ContainsKey` once the add has already thrown — free, because it only runs on the failing path.

## Decisions

**Caught, not pre-checked.** A duplicate is malformed input, so the cost belongs on that path rather than on a second hash lookup every well-formed pair would pay — and `Dictionary.TryAdd` is unavailable on netstandard2.0.

**This changes the exception, not the outcome.** Duplicates are still rejected. Letting the last occurrence win would be a data-integrity decision rather than a contract fix, and a silent one; if you would rather have that, it is a one-line change and its own conversation.

**Duplicate members of a POCO are untouched** and still take the last value — a member is matched by name rather than inserted into a dictionary, so it was never the same code path. Pinned by a test so the distinction is deliberate.

## Tests

Seven, five failing without the change. The other two pin what does *not* change: distinct keys that differ only by type (integer `1` and text `"1"` are different keys, and both must survive), and POCO members.

1111 tests pass on net10.0, up from 1104. All four TFMs build.

## One thing I found and did not change

`62 C3 28` — a text string whose payload is not valid UTF-8 — decodes without error to `U+FFFD U+0028`, because `Encoding.UTF8.GetString` substitutes rather than throws. RFC 8949 §5.5 is about text strings being valid UTF-8, so silently substituting is a data-integrity question for anyone treating decoded text as authentic.

That is a behaviour change rather than a contract fix, and it would want a decision — reject, or keep substituting and document it — so I have left it alone. Happy to take it if you want it, either way.

### DIFF
--- a/src/Dahomey.Cbor.Tests/DuplicateMapKeyTests.cs
+++ b/src/Dahomey.Cbor.Tests/DuplicateMapKeyTests.cs
@@ -1,5 +1,8 @@
+using Dahomey.Cbor.Attributes;
 using Dahomey.Cbor.ObjectModel;
+using Dahomey.Cbor.Serialization;
 using Dahomey.Cbor.Tests.Extensions;
+using Dahomey.Cbor.Util;
 using System.Collections.Generic;
 using Xunit;
 
@@ -109,6 +112,97 @@ namespace Dahomey.Cbor.Tests
         public class DuplicateHolder
         {
             public int A { get; set; }
+        }
+
+        /// <summary>
+        /// A type with a creator mapping takes a different branch of <c>ObjectConverter</c>: member
+        /// values are collected into dictionaries until the constructor can be called, rather than
+        /// being set on an instance. Those dictionaries reject a duplicate too, so the same contract
+        /// applies — a record or a <c>[CborConstructor]</c> type is an ordinary decode target.
+        /// </summary>
+        [Fact]
+        public void ADuplicateMemberOfAConstructedTypeThrowsCborException()
+        {
+            // a2 624964 0c 624964 0d  -- {"Id": 12, "Id": 13}
+            CborException exception = Assert.Throws<CborException>(
+                () => Cbor.Deserialize<ConstructedHolder>("A26249640C6249640D".HexToBytes()));
+
+            Assert.Contains("Duplicate", exception.Message);
+        }
+
+        /// <summary>The same type in <c>IntKeyMap</c> format, which collects by index instead.</summary>
+        [Fact]
+        public void ADuplicateIndexedMemberOfAConstructedTypeThrowsCborException()
+        {
+            // a2 00 0c 00 0d  -- {0: 12, 0: 13}
+            CborOptions options = new CborOptions { ObjectFormat = CborObjectFormat.IntKeyMap };
+
+            Assert.Throws<CborException>(
+                () => Cbor.Deserialize<IndexedConstructedHolder>("A2000C000D".HexToBytes(), options));
+        }
+
+        public class ConstructedHolder
+        {
+            public int Id { get; set; }
+
+            [CborConstructor]
+            public ConstructedHolder(int id)
+            {
+                Id = id;
+            }
+        }
+
+        public class IndexedConstructedHolder
+        {
+            [CborProperty(0)]
+            public int Id { get; set; }
+
+            [CborConstructor]
+            public IndexedConstructedHolder(int id)
+            {
+                Id = id;
+            }
+        }
+
+        /// <summary>
+        /// A null key is not a duplicate, and must not be reported as one. <c>ArgumentNullException</c>
+        /// derives from <see cref="System.ArgumentException"/>, so a catch that does not distinguish
+        /// them turns "the key was null" into "duplicate map key: " with nothing after the colon.
+        /// </summary>
+        [Fact]
+        public void ANullKeyIsReportedAsANullKeyRatherThanADuplicate()
+        {
+            // a1 f6 01  -- {null: 1}
+            CborException exception = Assert.Throws<CborException>(
+                () => Cbor.Deserialize<Dictionary<string, int>>("A1F601".HexToBytes()));
+
+            Assert.Contains("null", exception.Message);
+            Assert.DoesNotContain("Duplicate", exception.Message);
+        }
+
+        /// <summary>
+        /// A key from an untrusted document is not repeated at length into the exception message, which
+        /// typically reaches a log. The message says what was wrong without carrying a megabyte of
+        /// attacker-chosen text with it.
+        /// </summary>
+        [Fact]
+        public void ALongDuplicateKeyIsTruncatedInTheMessage()
+        {
+            string key = new string('k', 500);
+            byte[] keyBytes = System.Text.Encoding.UTF8.GetBytes(key);
+            ByteBufferWriter writer = new ByteBufferWriter();
+            CborWriter cborWriter = new CborWriter(writer);
+            cborWriter.WriteBeginMap(2);
+            cborWriter.WriteString(key);
+            cborWriter.WriteInt32(1);
+            cborWriter.WriteString(key);
+            cborWriter.WriteInt32(2);
+
+            CborException exception = Assert.Throws<CborException>(
+                () => Cbor.Deserialize<Dictionary<string, int>>(writer.WrittenSpan.ToArray()));
+
+            Assert.True(exception.Message.Length < 200, $"message was {exception.Message.Length} chars");
+            Assert.Contains("Duplicate map key", exception.Message);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


mcatanzariti (commented):
## Review

The diagnosis is right and the fix is in the right place. I checked that no `catch` upstream in `Cbor.cs` was converting this `ArgumentException` — it really did reach the caller as-is. Both read paths are genuinely covered: `ImmutableDictionaryConverter`, `InterfaceDictionaryConverter` and `DictionaryConverter` all derive from `AbstractDictionaryConverter`, so the single change does cover all five shapes as claimed. And it continues the thread started in #164, where `Issues/Issue0163.cs:66` already documented this exact collision as a hole in the contract.

Two things I'd want addressed before merge, and a few optional ones.

**Note on evidence:** `dotnet` is not available in my environment, so everything below is static reading of the code rather than an executed repro. The line numbers and branch conditions are explicit, but §1 in particular deserves a test to confirm.

---

### 1. The survey has a blind spot: POCOs with a constructor (`[CborConstructor]` / records)

The description says **"Every other one already threw `CborException`"** and that duplicate POCO members are untouched. That holds for POCOs with settable properties — but not for the *creator* path. In `ObjectConverter.cs`:

```csharp
context.creatorValues.Add(new RawString(memberName), value);   // line 686
context.regularValues!.Add(new RawString(memberName), value);  // line 690
context.creatorValuesByIndex.Add(memberIndex, value);          // lines 714, 739
context.regularValuesByIndex!.Add(memberIndex, value);         // lines 718, 743
```

Those are `Dictionary<RawString, object>` / `Dictionary<int, object>` (declared lines 45-48), and these branches are only taken when `_objectMapping.CreatorMapping != null`. So for a type like `ObjectWithConstructor2` in `CreatorMappingTests.cs`, the document `A2 6249640C 6249640D` (`{"Id":12,"Id":13}`) still throws a bare **`ArgumentException`** — the very escape this PR is closing. Same for the `IntKeyMap` format.

`DuplicatePocoMembersStillTakeTheLastValue` doesn't catch this because its `DuplicateHolder { public int A { get; set; } }` has no creator mapping and takes the other branch entirely.

Either give those four `Add` calls the same treatment plus a creator-mapping test, or leave it to a follow-up — but in that case the "every other one already threw `CborException`" claim in the description needs softening, since a record or constructor-based POCO is a very ordinary decode target.

### 2. `catch (ArgumentException)` also catches `ArgumentNullException`, mislabelling a null key

`ArgumentNullException` derives from `ArgumentException`, and `StringConverter.Read` returns `null` for a null token (`CborReader.cs:419-424`). So `A1 F6 01` deserialized into `Dictionary<string, int>` hits `Add(null, 1)` → `ArgumentNullException` → now caught and rethrown as **`[3] Duplicate map key: `** with an empty key, when there is no duplicate at all.

It's still a net improvement on the contract — that path was outside `CborException` before too — but the message is wrong. One line:

```csharp
catch (ArgumentException) when (key is not null)
```

with a dedicated message for the null key, or an explicit `key is null` check before the `Add`.

### 3. The catch can mislabel an exception from a caller-supplied `IDictionary`

`DictionaryConverter.InstantiateTempCollection()` does `new TC()` where `TC` is the caller's type. A custom `IDictionary<,>` whose `Add` throws `ArgumentException` for its own reasons (value validation, key domain checks) will be reported as "Duplicate map key". Niche, but the message asserts something untrue. A confirming `ContainsKey` inside the `catch` block is on the cold path, so it costs nothing on well-formed input.

### 4. The reported position isn't the key's

`BuildException` uses `_currentPos` at throw time, i.e. *after* the value has been read. For `A2616101616102` the message is `[7] …` while the offending key starts at offset 4 — `[7]` is the end of the document. This is consistent with every other exception in the reader, so it's not a defect in the code; but "names neither the key's position" in the description implies the new message does. Wording only.

### 5. Minor

- **Perf:** a `try/catch` region in `ReadMapItem` blocks JIT inlining of a method called once per pair on the hot read path. Probably negligible, but for a library that advertises high performance the alternative exists: `#if !NETSTANDARD2_0` + `Dictionary.TryAdd` for the concrete case, keeping the `catch` as the fallback for arbitrary `IDictionary`.
- **Hostile input in the message:** `{key}` interpolates an attacker-controlled, untruncated key into the exception message — a 1 MB text key ends up in the string and likely in logs. Truncating to ~64 chars seems prudent given that decoding untrusted frames is the whole argument for the change.
- The two paths produce slightly different messages for the same document (`Duplicate map key: "a"` via `CborValue.ToString()`, `Duplicate map key: a` via `string`). Cosmetic.

---

### Tests

Seven, well targeted and well documented — root, nested, integer key, both converters, and two pinning what must *not* change. Two gaps relative to the above: the constructor-based POCO (§1) and the null key (§2). The naming style (long sentences) differs from the older files (`FromObject`, `ConstructorByApi`) though it matches your recent ones — a convention call for the maintainer.

### On the UTF-8 note

Flagging `62 C3 28` without changing it is the right call: that's a behaviour change rather than a contract fix, and RFC 8949 §5.5 makes it a separate conversation. Agreed it should be its own PR.

### Verdict

Correct, well-motivated, low-risk, and it changes only the exception type — not the outcome. **Mergeable with §1 and §2 addressed**; §3-§5 are optional.

---

mcatanzariti (approved):
## Approving

All of it addressed, and §1 turned out wider than I flagged — six `Add` sites across `StringKeyMap`, `IntKeyMap` and the indexed variant, not the four I counted from line numbers, all now routed through one `AddMemberValue`. Writing the failing test first and correcting the description rather than quietly patching the code is the right way round.

The `ContainsKey` probe inside the `catch` is exactly the shape I had in mind, and folding the three messages into `MapKeyErrors` also removes the wording divergence I'd listed as cosmetic in §5 — good catch on your side.

On the perf point: thanks for measuring instead of taking "probably negligible" at face value. Your reasoning for not adding the `#if NETSTANDARD2_0` split is correct — `TryAdd` is on `Dictionary<K,V>` and not on the `IDictionary<K,V>` the converter holds, so it would buy an unmeasurable saving at the cost of a type test on every pair. Leave it as is.

### What I checked on this head

- **The sweep is complete now.** The only remaining `Add` calls on the read path are `readMembers.Add(...)`, and `readMembers` is a `HashSet<IMemberConverter>` (`ObjectConverter.cs:49`) whose `Add` returns a `bool` and never throws. No unguarded `Dictionary.Add` is left while decoding.
- `RawString.ToString()` (`Util/RawString.cs:58`) decodes the UTF-8, which is why the creator path reports `Duplicate map key: Id` and not a type name.
- `RawString` is a struct, so omitting the null branch in `AddMemberValue` is correct rather than an oversight.

### Minor, none blocking, none needed for this PR

1. `CborException` only has a `(string)` constructor, so the `Rejected` branch keeps the underlying exception's *message* but not the exception. That's the existing API, not something this PR introduced — a `CborException(string, Exception)` overload would be a public API addition and its own change.
2. Theoretical residue: if a caller-supplied `IDictionary` has a `ContainsKey` that throws, that exception escapes from the `catch` block and replaces the `CborException`. Lab case, not worth code.
3. Unverified, worth a glance at the build output: `AddMemberValue` takes `Dictionary<TKey, object?>` while the fields are `Dictionary<TKey, object>`, which may raise CS8620 under `<Nullable>enable</Nullable>`. Nothing is warnings-as-errors here and the original code already passed an `object?` into a `Dictionary<,object>.Add`, so it builds either way. (I still have no `dotnet` available, so all of the above remains static reading.)

### Before merge

`build_and_test` has never run against this PR head — `97c15e5` currently carries only the GitGuardian check, since the push landed inside the Actions outage. Your fork run is good evidence and the `src/` tree being byte-identical is checkable, but now that Actions is back I'd rather have the workflow green on the PR itself. I'll merge once it is.

---

PR #206, opened 2026-08-12
Give decimal an interoperable encoding, closes #192

mcatanzariti:
Closes #192, taking option 2 from the issue: opt-in on write, reader widened immediately and independently.

## Write: `CborOptions.DecimalFormat`

`DecimalFloat` stays the default, so no existing document moves. `DecimalFraction` writes the RFC 8949 §3.4.4 form — tag 4 over `[-scale, mantissa]`, which is what a `decimal` already is, so the conversion is total and needs no range policy.

```
273.15m            -> C4 82 21 19 6AB3                            (was FC 0000000000006AB3 0000000000020000)
decimal.MaxValue   -> C4 82 00 C2 4C FFFFFFFFFFFFFFFFFFFFFFFF     (mantissa past 2^64 takes the bignum tag)
0.00m / 0m         -> C4 82 21 00 / C4 82 00 00                   (the scale survives, as it does today)
```

The choice is made in `DecimalConverter` rather than in `CborWriter`, and that placement is load-bearing: `DeterministicKeyOrder` encodes keys through the key converter, so the bytes it sorts are the bytes that go out. A flag on the writer would have sorted one form and emitted the other. `CborWriter.WriteDecimal(value)` is unchanged — it holds no options — and a second overload takes the format for custom converters.

## Read: unconditional

`ReadDecimal` walks the tag stack the way `ReadBigInteger` does instead of calling `SkipSemanticTag`, because here the tag carries the meaning. Tag 4 was a `CborException` before, so accepting it takes nothing away, and a service can turn the setting on without losing the documents it wrote before.

Reading is the more generous half: an indefinite-length pair and an unnormalised mantissa are both accepted, so `[-30, 100]` reads as `1E-28`. What a `decimal` genuinely cannot hold — a mantissa wider than 96 bits, or a scale that will not reduce to 28 — is a `CborException` rather than a rounded value. The exponent is bounded before `BigInteger.Pow` sees it, and the reduction divides once against a divisor bounded by `Log10` of the mantissa, so neither is driven by a number a hostile document merely claims.

## Source generation and CDDL

`DecimalFormat` is declarable on `[CborSourceGenerationOptions]`, and it is the only encoding of a `decimal` a schema can describe: with it, such a member renders as `#6.4([int, (int / #6.2(bstr) / #6.3(bstr))])` instead of raising CBOR1011. The mantissa is the same three-way choice `BigInteger` renders as, for the same reason.

## Two things worth a second opinion

- **Minor source break.** `DecimalConverter` now takes `CborOptions`, matching `DateTimeConverter`. Anyone calling `new DecimalConverter()` has to pass options.
- **The object model is asymmetric.** A `CborDecimal` writes as tag 4 like any other decimal, but reading those bytes back gives a `CborArray` tagged 4 — the DOM has no type for a decimal fraction, and giving it one is #170's business. The document is right either way; only the DOM type differs from what wrote it. Documented in the README next to the equivalent typed-array caveat, and pinned by a test that asserts exactly this.

## Tests

`DecimalFractionTests` covers both directions, the tag stack, indefinite length, the out-of-range and malformed rejections, dictionary keys under `Deterministic` (the pair sorts one way as decimal fractions and the other way as `DecimalFloat`), the DOM asymmetry, and negative zero losing its sign. `Cddl/CddlDecimalFractionTests` pins the rendering and validates real output against the schema. Both new fixture types are enrolled in `GeneratedCorpusTests`, which was missing `DecimalFormat` from the settings it copies onto the reflection side — without that the corpus would have compared tag 4 against the `FC` form.

1726 tests green on net10.0 and net8.0, 35 on the generator, netstandard2.0 compiles. net9.0 has no runtime available in the container, and the CDDL cases skip locally without the `cddl` gem — both are on CI.

https://claude.ai/code/session_01SnnTU1dzbhPdmUQBs8rmw9

---
_Generated by [Claude Code](https://claude.ai/code/session_01SnnTU1dzbhPdmUQBs8rmw9)_

### DIFF
--- a/src/Dahomey.Cbor/CborOptions.cs
+++ b/src/Dahomey.Cbor/CborOptions.cs
@@ -135,10 +135,17 @@ namespace Dahomey.Cbor
         /// decimal.
         /// </summary>
         /// <remarks>
-        /// Lossless in both directions and no wider than the type. A <see cref="decimal"/> of scale
-        /// <c>s</c> and mantissa <c>m</c> is exactly the decimal fraction <c>[-s, m]</c>, so the
-        /// scale survives the round trip - <c>0.00m</c> writes as <c>[-2, 0]</c> and <c>0m</c> as
-        /// <c>[0, 0]</c>, the same distinction <see cref="DecimalFloat"/> keeps.
+        /// Carries every value the type holds, and its scale with it: a <see cref="decimal"/> of scale
+        /// <c>s</c> and mantissa <c>m</c> is exactly the decimal fraction <c>[-s, m]</c>, so
+        /// <c>0.00m</c> writes as <c>[-2, 0]</c> and <c>0m</c> as <c>[0, 0]</c> - the same distinction
+        /// <see cref="DecimalFloat"/> keeps.
+        /// <para>
+        /// One exception, and it is in the encoding rather than in this library: a decimal fraction has
+        /// no signed zero, so <c>-0.00m</c> reads back as <c>0.00m</c>. Equal by every comparison the
+        /// language offers, and distinguishable only by <see cref="decimal.GetBits(decimal)"/> or by
+        /// rendering it - where <see cref="DecimalFloat"/>, which stores the sign bit as it stands,
+        /// keeps it.
+        /// </para>
         /// </remarks>
         DecimalFraction = 1,
     }
--- a/src/Dahomey.Cbor/CborOptions.cs
+++ b/src/Dahomey.Cbor/CborOptions.cs
@@ -101,6 +101,48 @@ namespace Dahomey.Cbor
         ReadWriteLittleEndian = Read | WriteLittleEndian,
     }
 
+    /// <summary>
+    /// Which of the two encodings a <see cref="decimal"/> is written as.
+    /// </summary>
+    /// <remarks>
+    /// Reading is not affected and needs no setting: a <see cref="decimal"/> is read from either form
+    /// whatever this says, so widening a service's input costs nothing and turning this on does not
+    /// stop it reading the documents it wrote before.
+    /// <para>
+    /// Only writing has to choose, because the choice moves bytes: every <see cref="decimal"/> this
+    /// library has ever written uses <see cref="DecimalFloat"/>, so that stays the default and no
+    /// existing document changes shape underneath a caller who has not asked for it.
+    /// </para>
+    /// </remarks>
+    public enum DecimalFormat
+    {
+        /// <summary>
+        /// Major type 7 with additional information 28, followed by the sixteen raw bytes of the
+        /// value. The historical form, and the default.
+        /// </summary>
+        /// <remarks>
+        /// RFC 8949 §3.3 lists additional information 28-30 in major type 7 as reserved, so this
+        /// occupies a slot the format has not assigned: it round-trips through this library and no
+        /// other decoder reads it. Keep it for a contract whose only participants are Dahomey.Cbor,
+        /// or while documents written by an earlier version are still in circulation; choose
+        /// <see cref="DecimalFraction"/> for anything read outside this library.
+        /// </remarks>
+        DecimalFloat = 0,
+
+        /// <summary>
+        /// The RFC 8949 §3.4.4 decimal fraction: tag 4 over the two-element array
+        /// <c>[exponent, mantissa]</c>, which is what every other CBOR implementation means by a
+        /// decimal.
+        /// </summary>
+        /// <remarks>
+        /// Lossless in both directions and no wider than the type. A <see cref="decimal"/> of scale
+        /// <c>s</c> and mantissa <c>m</c> is exactly the decimal fraction <c>[-s, m]</c>, so the
+        /// scale survives the round trip - <c>0.00m</c> writes as <c>[-2, 0]</c> and <c>0m</c> as
+        /// <c>[0, 0]</c>, the same distinction <see cref="DecimalFloat"/> keeps.
+        /// </remarks>
+        DecimalFraction = 1,
+    }
+
     public class CborOptions
     {
         public static CborOptions Default { get; } = new CborOptions()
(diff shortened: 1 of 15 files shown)


src/Dahomey.Cbor/Serialization/CborReader.cs
```diff
@@ -680,6 +719,153 @@ public decimal ReadDecimal()
...
+                        ThrowDecimalFractionOutOfRange(exponent, mantissa);
+                    }
+
+                    scale = MAX_DECIMAL_SCALE;
+                }
+            }
+
+            // 96 bits, checked after the scale reduction above rather than before it, since dividing
+            // out the trailing zeros is what brings a value such as [-30, 10^29] into range.
+            if (!(magnitude >> 96).IsZero)
```
mcatanzariti, line 866: **Representable values are rejected when the mantissa is unnormalised and the scale is not.**

The trailing factors of ten are divided out only in the `scale > MAX_DECIMAL_SCALE` branch. When the scale is already inside 0..28 there is no reduction at all, so a mantissa wider than 96 bits reaches this check unreduced and is refused even though the value it denotes fits `decimal` comfortably. And when the scale *is* too large, the reduction stops the moment `scale` reaches 28 rather than continuing while the magnitude still overflows 96 bits and is still divisible by ten.

Verified against the built library on net8.0 — all three of these throw `CborException: ... is not exactly representable as a decimal`:

| bytes | decimal fraction | value | actual |
|---|---|---|---|
| `C48220C24D01431E0FAE6D7217CAA0000000` | `[-1, 10^29]` | `1E+28` | throws |
| `C48221C24D0C9F2C9CD04674EDEA40000000` | `[-2, 10^30]` | `1E+28` | throws |
| `C4823827C2520125DFA371A19E6F7CB54395CA0000000000` | `[-40, 10^41]` | `10` | throws |

`1E+28` is well inside `decimal.MaxValue` (7.92E+28), and `10` obviously is. This is exactly the shape the method's own remark promises to handle — *"A mantissa carrying trailing zeros is not out of range just because its exponent is: the factors of ten are divided out first ... Producers outside .NET have no reason to normalize."* A `java.math.BigDecimal` with unscaled value `10^29` and scale 1 encodes as row 1, so this is not a synthetic case.

The reduction wants to be driven by what does not fit rather than by the scale alone: after the existing scale handling, keep dividing by ten (lowering `scale`, floor 0) while `(magnitude >> 96)` is non-zero and `magnitude % 10 == 0`, and only then declare it out of range.

---

src/Dahomey.Cbor/Serialization/CborReader.cs
```diff
@@ -680,6 +719,153 @@ public decimal ReadDecimal()
...
+                    // digit count without materialising anything: this keeps Pow's argument tied to the
+                    // size of the mantissa actually in the document rather than to the exponent it
+                    // claims.
+                    if (excess > (long)BigInteger.Log10(magnitude) + 1)
+                    {
+                        ThrowDecimalFractionOutOfRange(exponent, mantissa);
+                    }
+
+                    magnitude = BigInteger.DivRem(
+                        magnitude, BigInteger.Pow(10, (int)excess), out BigInteger remainder);
```
mcatanzariti: **`Pow`'s argument is bounded by the document's size, not by anything that could fit.**

The `Log10` guard above ties `excess` to the mantissa's digit count, which the comment presents as the win — but the digit count is attacker-chosen and linear in the bytes sent, while `Pow` plus `DivRem` on numbers that size is superlinear. A mantissa that is a tag‑2 bignum of `N` bytes (~2.4·`N` digits) with exponent `-(digits + 28)` sails through the guard and then materialises `BigInteger.Pow(10, ~2.4N)`.

Measured on net8.0 (`Cbor.Deserialize<decimal>`), each case ending in a throw or `1E-28`:

| input | mantissa digits | elapsed |
|---|---|---|
| 8 KB | 20 000 | 1 ms |
| 16 KB | 40 000 | 4 ms |
| 33 KB | 80 000 | 14 ms |

Roughly quadratic, so a ~1 MB `decimal` member is seconds of CPU per request for a service that accepts CBOR off the network — a large amplification for a value that cannot possibly be representable. The information needed to reject it early is already here: for the result to fit, `magnitude` must be below `2^96 · 10^excess`, so a bit-length pre-filter (`magnitude.GetBitLength()` vs `96 + excess * log2(10)`) rejects the hostile shape before any `Pow`, and leaves the legitimate unnormalised case — where `excess` is small — untouched.

---

src/Dahomey.Cbor/CborOptions.cs
```diff
@@ -101,6 +101,48 @@ public enum TypedArrayMode
...
+        /// </remarks>
+        DecimalFloat = 0,
+
+        /// <summary>
+        /// The RFC 8949 §3.4.4 decimal fraction: tag 4 over the two-element array
+        /// <c>[exponent, mantissa]</c>, which is what every other CBOR implementation means by a
+        /// decimal.
+        /// </summary>
+        /// <remarks>
+        /// Lossless in both directions and no wider than the type. A <see cref="decimal"/> of scale
```
mcatanzariti: **"Lossless in both directions" has one exception, and it is not documented where a caller would look.**

`DecimalFractionTests.NegativeZeroLosesItsSign` pins that `-0.00m` writes as `C4 82 21 00` and reads back as `+0.00m` — tag 4 has no sign-of-zero. That is a defensible trade, but this doc and the README (`§Decimals`: *"The conversion is total and lossless in both directions, with no range or precision policy to pick"*) both assert the opposite, so a caller who stores `-0m` and later compares `decimal.GetBits` or renders with a sign-showing format finds a difference the documentation told them could not happen.

Worth a clause here and in the README, in the same spirit as the object-model asymmetry the README already calls out — the test's own summary is already the right wording.

---

src/Dahomey.Cbor/Serialization/Converters/DecimalConverter.cs
```diff
@@ -2,14 +2,25 @@
 {
     public class DecimalConverter : CborConverterBase<decimal>
     {
+        private readonly CborOptions _options;
+
+        public DecimalConverter(CborOptions options)
```
mcatanzariti, line 7: **Source-breaking: the implicit public parameterless constructor is gone.**

`DecimalConverter` is public, and adding this constructor removes the compiler-supplied one. Existing consumer code that swaps a primitive converter by hand —

```csharp
options.Registry.ConverterRegistry.RegisterConverter(typeof(decimal), new DecimalConverter());
```

— stops compiling with CS7036 on upgrade. `DateTimeConverter` has always taken options so the resulting shape is consistent, but this type's contract changed. Either keep a parameterless overload chaining to `CborOptions.Default`, or note the break in the release notes so it is not discovered at build time.

---

mcatanzariti (commented):
Reviewed the decimal-fraction work end to end (reader, writer, converters, generator, CDDL, docs). The design is sound and the whole suite passes on net8.0 (1726 passed), but one reader case rejects documents it can read.

**One correctness bug.** `DecimalFromDecimalFraction` divides trailing factors of ten out of the mantissa only when the *scale* is out of range, never when the *mantissa* is. So `[-1, 10^29]` — which is `1E+28`, comfortably inside `decimal.MaxValue` — throws, as do `[-2, 10^30]` and `[-40, 10^41]`. Confirmed against the built library; details inline. This is precisely the unnormalised shape the method's remark promises to accept, and it is what `java.math.BigDecimal` emits when its unscaled value carries the zeros.

**One hardening note.** The `Log10` guard bounds `BigInteger.Pow`'s exponent by the mantissa's digit count, i.e. by the attacker's byte count, and the arithmetic on numbers that size is superlinear: 33 KB of input measures 14 ms, so ~1 MB is seconds of CPU for a value that cannot fit. A bit-length pre-filter rejects it before any `Pow`.

Two smaller ones: the "lossless in both directions" claim (doc + README) has a negative-zero exception the PR's own test pins but neither doc mentions, and `DecimalConverter` silently loses its public parameterless constructor.

Everything else held up under scrutiny — the write side's scale/sign extraction, `WriteBigInteger` reaching for the bignum tag only past 2^64, the read side's tag-stack walk mirroring `ReadBigInteger`, the indefinite-length array acceptance, zero-at-any-exponent, the attribute→emitted-options round trip, and the `WritesDecimalFractions`/CBOR1011 coupling in the CDDL path. The corpus test carrying `DecimalFormat` onto the reflection side is the right call, and it does bind.

---

PR #208, opened 2026-08-12
Make every CDDL rule name a legal RFC 8610 id, and unique

Rafael-SOWNet:
The two points your #159 review left open, both of which you said needed a design call. **1723 library tests plus 35 generator tests**, +9 here, green on net10.0; three TFMs build.

## Point 6 — a non-ASCII type name

Verified rather than inferred: it is a hard failure, not a schema a tool merely dislikes.

```
Größe = { "v" : int }
```

```
*** Look for syntax problems around the %%% markers:
start = Gr%%%öße
*** Parse error at 10 upto 10 of 36 (1382).
exit=65
```

And a schema that does not parse cannot fail an instance check, so nothing downstream reports it either — the failure is silent everywhere except at the gem's own exit code.

Rule names are now folded onto `id = EALPHA *(*("-" / ".") (EALPHA / DIGIT))`: each character outside it becomes `_` followed by its code point in uppercase hex, a surrogate pair escaped as the one code point it is. `Größe` → `Gr_00F6_00DFe`.

Escaping rather than dropping, because dropping collapses `Café` and `Cafe` onto one rule — which is the whole subject of the other half.

## Point 5 — the collision-breaking name colliding

**The collision you describe is not constructible within one compilation**, and that is worth stating before the fix. It needs a namespace chain and a containing-type chain that flatten to the same string, and C# refuses the pair outright:

```csharp
namespace A.B { public class C { } }
namespace A { public class B { public class C { } } }
```

```
error CS0101: The namespace 'A' already contains a definition for 'B'
```

Two assemblies can each supply one half, so it is reachable across an assembly boundary — but there both types render as `global::A.B.C` under `FullyQualifiedFormat`, which is the key `CddlEmitter.byKey`, `ruleNames` and `TypeCollector` all identify a type by, so the rule name is not the first thing that breaks. I have not built that case; I am noting it rather than claiming it, and it is not what this PR fixes.

So the fix is not a better rendering — no rendering scheme can be shown to be injective over an input C# cannot produce. It is a guarantee: after escaping, names that still collide are separated by ordinal order of the full type key, the first keeping the bare name and the rest taking `-2`, `-3`. Ordinal order of the key rather than arrival order, because a schema is compared against other copies of itself, so the mapping has to be a function of the set of types alone. That is the property a rule name needs and `Emitter.UniqueAccessorName` deliberately does not.

## What makes the second half load-bearing rather than defensive

Escaping folds a character outside ASCII onto a `_` sequence an identifier is free to spell out, so `Café` and `Caf_00E9` arrive asking for one rule name — and that pair *is* constructible, which is why it is the fixture.

And a duplicate rule name is the one malformation an emitted schema can carry with nothing complaining. Measured, not assumed: with the uniqueness pass disabled and escaping left in, a schema carrying two `Caf_00E9` rules still passes `EveryEmittedSchemaParses` — the gem reads the second definition as a silent shadow of the first. A schema whose rules collide describes something other than the types it was generated from, and every instance check run against it then passes or fails on the surviving rule.

## Tests, and that they bind

`CddlIdentifierSyntaxTests` declares `Größe`, `Café` and `Caf_00E9`. All three are enrolled in `GeneratedCorpusTests` (six cases) as well, so their generated converters stay pinned to the reflection path's bytes, and `CddlSchemaParsesTests` picks the context up by assembly scan.

Each half was checked by disabling it and re-running, which is also where the two measurements above come from:

| disabled | result |
|---|---|
| escaping | `TheEscapedAndSuffixedSchemaParses` and `EveryEmittedSchemaParses` both fail — `*** Parse error at 0 upto 679 of 826` |
| the uniqueness pass | `TwoTypesAskingForOneRuleNameEachGetTheirOwn` fails on the missing `-2`, and every gem check stays green |

## One thing I left out deliberately

`PolymorphicShape` mints `{ruleName}-poly` from a rule name, and `TypeNames` does not reserve that derived form — so a type whose own rule name is another type's plus `-poly` would collide with it. Reaching it needs a namespace `X` holding a type named `poly`, alongside a polymorphic type whose rule name is `X`, with the first in a collision bucket so it takes the qualified form at all. I have not built it and am not claiming it is reachable. Reserving the derived names would mean `TypeNames` knowing the suffix `CddlPolymorphism` appends, which is a coupling I would rather you agree to than assume — say the word and it is a two-line change.

---

### DIFF
--- a/src/Dahomey.Cbor.Generator/CddlEmitter.cs
+++ b/src/Dahomey.Cbor.Generator/CddlEmitter.cs
@@ -30,9 +30,12 @@ namespace Dahomey.Cbor.Generator
                 byKey[model.Symbol.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat)] = model;
             }
 
-            IReadOnlyDictionary<string, string> ruleNames = TypeNames.BuildRuleNames(ordered);
-            Dictionary<string, PolymorphicShape> shapes =
-                CddlPolymorphism.BuildShapes(ordered, byKey, ruleNames);
+            // Shapes first, because a polymorphic type is emitted under a second rule name derived from
+            // its first, and the uniqueness pass has to reserve both together. Only the order of a
+            // choice's arms depends on the names, so that one step waits.
+            Dictionary<string, PolymorphicShape> shapes = CddlPolymorphism.BuildShapes(ordered, byKey);
+            IReadOnlyDictionary<string, string> ruleNames = TypeNames.BuildRuleNames(ordered, shapes);
+            CddlPolymorphism.SortArms(shapes, ruleNames);
 
             // Use sites resolve through their own name table: a member declared as a base type names
             // that base's `-poly` rule, a member declared as a leaf names the leaf's bare rule. Keeping
--- a/src/Dahomey.Cbor.Generator/CddlPolymorphism.cs
+++ b/src/Dahomey.Cbor.Generator/CddlPolymorphism.cs
@@ -70,6 +70,27 @@ namespace Dahomey.Cbor.Generator
         {
             return EmitsPoly ? ruleName + "-poly" : ruleName;
         }
+
+        /// <summary>
+        /// The second rule name this type occupies, or null when it is emitted under one name only.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="ArmName"/> and <see cref="ReferenceName"/> mint this from a rule name that has
+        /// already been handed out, so it is never a candidate the uniqueness pass sees -- which is why
+        /// <see cref="TypeNames.BuildRuleNames"/> asks for it and reserves the two names together. A
+        /// type whose own declaration name reads <c>X-poly</c> would otherwise be free to take the
+        /// derived form of a polymorphic <c>X</c>, and two rules would be emitted under one name.
+        /// <para>
+        /// An over-approximation by one case: a type reported by CBOR1012 is skipped before either rule
+        /// is emitted, so its reservation holds a name nothing uses. That costs another type a suffix in
+        /// a compilation that already carries an error, which is cheaper than making the reservation
+        /// depend on diagnostics that have not run yet.
+        /// </para>
+        /// </remarks>
+        public string? DerivedRuleName(string ruleName)
+        {
+            return EmitsPoly ? ArmName(ruleName) : null;
+        }
     }
 
     /// <summary>
@@ -78,10 +99,15 @@ namespace Dahomey.Cbor.Generator
     /// </summary>
     internal static class CddlPolymorphism
     {
+        /// <remarks>
+        /// Takes no rule names, and that is what makes the ordering work: <see cref="TypeNames.BuildRuleNames"/>
+        /// needs to know which types emit a <c>-poly</c> rule *before* it hands names out, so the one
+        /// thing here that does depend on a name -- the order of a choice's arms -- is settled afterwards
+        /// by <see cref="SortArms"/>. Nothing else in a shape is derived from a rule name.
+        /// </remarks>
         public static Dictionary<string, PolymorphicShape> BuildShapes(
             IReadOnlyList<TypeModel> ordered,
-            IReadOnlyDictionary<string, TypeModel> byKey,
-            IReadOnlyDictionary<string, string> ruleNames)
+            IReadOnlyDictionary<string, TypeModel> byKey)
         {
             Dictionary<string, PolymorphicShape> shapes = new Dictionary<string, PolymorphicShape>();
 
@@ -119,11 +145,6 @@ namespace Dahomey.Cbor.Generator
 
             foreach (PolymorphicShape shape in shapes.Values)
             {
-                // Ordered by rule name rather than by collection order, so the emitted choice is a pure
-                // function of the source and the schema stays byte-stable across builds.
-                shape.Subtypes.Sort((left, right) => string.CompareOrdinal(
-                    ruleNames[Key(left.Symbol)], ruleNames[Key(right.Symbol)]));
-
                 shape.EmitsBare = shape.IsInstantiable
                     && !(shape.WritesDiscriminator && shape.Policy == "Always");
                 shape.EmitsPoly = shape.Subtypes.Count > 0
@@ -136,6 +157,21 @@ namespace Dahomey.Cbor.Generator
             }
 
             return shapes;
+        }
+
+        /// <summary>
+        /// Orders each type choice's arms by rule name rather than by collection order, so the emitted
+        /// choice is a pure function of the source and the schema stays byte-stable across builds.
+        /// </summary>
+        public static void SortArms(
+            IReadOnlyDictionary<string, PolymorphicShape> shapes,
+            IReadOnlyDictionary<string, string> ruleNames)
+        {
+            foreach (PolymorphicShape shape in shapes.Values)
+            {
+                shape.Subtypes.Sort((left, right) => string.CompareOrdinal(
+                    ruleNames[Key(left.Symbol)], ruleNames[Key(right.Symbol)]));
+            }
         }
 
         /// <summary>
--- a/src/Dahomey.Cbor.Generator/TypeNames.cs
+++ b/src/Dahomey.Cbor.Generator/TypeNames.cs
@@ -1,5 +1,7 @@
 using Microsoft.CodeAnalysis;
+using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text;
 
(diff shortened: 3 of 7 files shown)


mcatanzariti (commented):
Reviewed against a local checkout with the `cddl` gem present, so the `[CddlFact]` tests ran rather than skipped.

**Verification.** `CDDL_REQUIRED=1`, gem 0.12.14: **1723/1723 on net10.0, 1723/1723 on net8.0, 35/35 generator tests**, all four TFMs build. The numbers in the description are exact. Point 6 reproduces independently: `Größe = { "v" : int }` gives `*** Parse error` and `exit=65`.

**The mechanism holds up.** I checked the three properties `ResolveRemainingCollisions` claims. Injectivity comes from the `used` set and does not depend on the shape of the names, including the case where a contested group `X` mints `X-2` while another contested group is itself named `X-2`. The uncontested-first pass is genuinely necessary, and the ordering — ordinal on the candidate, then ordinal on the type key — makes the mapping a function of the type set alone, so the determinism argument is sound. Keeping `BuildRuleNames` (the only caller is `CddlEmitter.cs:33`) separate from `AccessorName`/`UniqueAccessorName` is right: escaped names contain `-` and are not legal C# identifiers, so they must not reach the accessor emitter.

Three things below. The first is a defect; the other two are about the evidence rather than the code.

---

## 1. The `-poly` case you left out is reachable, and I built it

Your description says reaching it needs "a namespace `X` holding a type named `poly`, alongside a polymorphic type whose rule name is `X`" and that you are not claiming it is reachable. That particular shape is indeed blocked — a namespace and a type of the same name conflict. But going through a **nested type** sidesteps the conflict entirely:

```csharp
namespace ProbeLeft {
    public class X {
        public int Id { get; set; }
        public class poly { public int Value { get; set; } }
    }
    [CborDiscriminator("sub")] public class XSub : X { public int Extra { get; set; } }
}
namespace ProbeRight {
    public class X { public int Id { get; set; } }
    public class poly { public int Value { get; set; } }
}
```

Five ordinary types, one compilation, no cross-assembly trick. `ProbeLeft.X` is contested with `ProbeRight.X` so it takes the qualified form; `ProbeLeft.X.poly` is contested with `ProbeRight.poly` so it takes a qualified form that includes the containing type — which is the base's qualified name plus `-poly`. Emitted schema:

```
Dahomey-Cbor-Tests-Cddl-ProbeLeft-X = {
  "Id": -2147483648..2147483647,
}

Dahomey-Cbor-Tests-Cddl-ProbeLeft-X-poly = Dahomey-Cbor-Tests-Cddl-ProbeLeft-X / XSub-poly

...

Dahomey-Cbor-Tests-Cddl-ProbeLeft-X-poly = {
  "Value": -2147483648..2147483647,
}
```

Two rules, one name. The uniqueness pass cannot see it because `PolymorphicShape.ArmName`/`ReferenceName` mint `-poly` *after* `BuildRuleNames` has finished. This is precisely the malformation the PR exists to eliminate, left open at a spot where it is constructible — so the two-line change you offered belongs in this PR rather than after it.

Note the coupling you were reluctant to introduce is smaller than reserving a literal suffix: `BuildRuleNames` could take the set of derived forms to reserve, or `ResolveRemainingCollisions` could seed `used` with `candidate + "-poly"` for every shape that emits one. Either keeps the suffix knowledge in `CddlPolymorphism` where it belongs.

## 2. "Nothing complains about a duplicate rule" is true only for an *identical* redefinition

This matters because it is the measurement the "load-bearing rather than defensive" argument rests on. Gem 0.12.14 distinguishes two cases:

| duplicate rule | behaviour |
|---|---|
| identical bodies | `*** Warning: Identical redefinition of ...` on stderr, **exit 0** |
| differing bodies | `RuntimeError: Duplicate rule definition ...`, **exit 1** |

Your disabled-uniqueness experiment stays green only because `Café` and `Caf_00E9` both happen to be `{ "Value": int }`. Give either one a different member and the same experiment fails loudly. The conclusion — that the uniqueness pass is needed — is still correct, but as written the evidence does not carry it, and the doc comment on `BuildRuleNames` ("the one malformation an emitted schema can carry without any tool complaining") should be narrowed to the identical case. The same sentence appears in `CddlRuleNamingTests`, inherited from #159.

Related, and pre-existing rather than yours: `CddlTool.Parse` only tolerates exit 0 and 65 and throws on anything else, so a differing-bodies duplicate surfaces as `cddl exited with 1` instead of a clean "did not parse" result.

## 3. `TwoTypesAskingForOneRuleNameEachGetTheirOwn` does not pin the property it documents

The test asserts that both `Caf_00E9` and `Caf_00E9-2` are present. It cannot assert *which type* holds the bare name, because both fixtures are `{ "Value": int }` and the two emitted rules are byte-identical. So if the tie-break flipped — the exact regression the ordinal-ordering design exists to prevent, and the one you argue hardest for in the description — the test would still pass.

Giving the two fixtures different member shapes fixes this and point 2 at the same time: the assertion becomes able to name the owner of the bare rule, and the disabled-uniqueness experiment turns into a hard gem error instead of a warning.

## Minor

- **The escape is not injective**, which the doc comment on `SanitizeRuleName` implies it is ("keeps two names differing only outside ASCII apart"). `X4` is not self-delimiting: `U+1F60` followed by a literal `0` and `U+1F600` both fold to `_1F600`. Harmless — the uniqueness pass catches it — but it is the `Café`/`Caf_00E9` reasoning applied a second time, so the comment is worth softening rather than leaving as a stronger claim than the code makes.
- **`Assert.DoesNotContain("ö", schema)`** covers the whole schema, not just rule names. CDDL text literals accept `NONASCII`, so adding a member named `Grüße` to `Größe` would fail the test with nothing wrong.
- **Readability asymmetry**, for the record rather than as a request: types colliding *before* escaping get the informative qualified form, types colliding only *after* get an opaque `-2`. Not fixable here — the qualified forms of `Café` and `Caf_00E9` in one namespace collide too — so the `-2` fallback is genuinely the right floor.
- **Stale comment, not yours:** `CddlRuleNamingTests` says declaring both nested types as roots "fails to compile (CS0102)". `Emitter.UniqueAccessorName` resolves that now; a context declaring two same-short-named roots compiles fine.

---

mcatanzariti (changes requested):
Formalising the state of my review above: **point 1 is blocking.**

The `-poly` collision is not a hypothetical you declined to build — it is constructible from five ordinary types in one compilation, via a nested type rather than the namespace/type pair that C# rejects, and it emits two rules under one name. That is the exact malformation this PR exists to eliminate, so shipping the uniqueness guarantee with that hole open would make the guarantee conditional in a way the doc comment on `BuildRuleNames` says it is not.

What I am asking for:

- Reserve the derived `-poly` forms so the uniqueness pass covers them. Seeding `used` in `ResolveRemainingCollisions` with the derived names keeps the suffix knowledge inside `CddlPolymorphism` and avoids the coupling you flagged as needing a design call.
- A fixture pinning it, along the lines of the repro in my review above.

Points 2 and 3 are not blocking on their own, but 3 is cheap and lands in the same place: giving the two `Café`/`Caf_00E9` fixtures different member shapes makes the tie-break assertion able to name which type keeps the bare rule, and turns the disabled-uniqueness experiment into a hard gem error instead of a warning that exits 0. Worth doing in the same pass.

Everything else — the escaping, the ordinal tie-break, the separation from `AccessorName` — I checked and it holds.

---

mcatanzariti (commented):
Re-reviewed at e91d189. All three points are addressed, and the blocking one is resolved.

**Verification.** Four TFMs build. With the gem present and `CDDL_REQUIRED=1`: **1735/1735 on net10.0, 1735/1735 on net8.0, 35/35 generator tests.** Your numbers again match exactly.

**The fixture binds.** I checked it the way you checked yours — `DerivedRuleName` stubbed to return `null`, everything else untouched. Three tests fail: `ATypeNamedForThePolySuffixDoesNotTakeAPolymorphicTypesSecondRule` on the missing `-2`, `TheSchemaCarryingBothParses`, and `EveryEmittedSchemaParses`. That last one is the payoff from giving the two bodies different members: the duplicate is now exit 1 rather than a warning at exit 0, so the assembly-wide net catches it instead of walking past it. Point 3 did real work for point 1.

**The restructuring is the right shape.** Building shapes first and deferring only the arm ordering to `SortArms` is a clean cut — the dependency was never on the names themselves except for that one sort, and separating it makes that explicit rather than incidental. Keeping the suffix literal in `PolymorphicShape.DerivedRuleName` and having `TypeNames` ask for it is better than what I suggested; `TypeNames` still doesn't know what the suffix is.

`Take` is correct. Reserving the pair atomically — retrying the whole candidate when either name is taken, rather than taking the bare name and discovering the derived one is gone — is what keeps a type's two rules from being split across a suffix, and recomputing `derived` per iteration is what makes that work. The unbounded loop terminates: each iteration tries a fresh string and `used` is finite.

I also checked the consequence you call out, that the uncontested pass can now suffix. Ordering it ordinally is necessary and sufficient for determinism there. The invariant the old comment claimed — that a suffix can never displace a name some other type was always going to hold — still holds, though now for a reason worth stating: a suffixed form ends in `-<digits>`, and no candidate can, since every segment joined by `-` is a C# identifier and none can be a bare number. `candidate-2` also can never equal some other type's derived `Y-poly`. So the displacement is unreachable rather than merely unlikely.

**One nit, non-blocking.** `RuleNames` treats any line containing `" = "` that doesn't start with a space as a rule. No header comment contains one today, so it is correct as written, but it would start reporting a comment as a rule name if one ever gained an `=`. Anchoring on a line that doesn't start with `;` would close it.

Nothing else from me. The escaping, the ordinal tie-break, the two-name reservation and the corrected doc comments all hold up.

---

mcatanzariti (approved):
Approving — this clears the changes I requested.

The `-poly` reservation closes the collision at its source rather than papering over it, and the verification is in my review above: four TFMs building, 1735/1735 on net10.0 and net8.0, 35/35 generator tests, and the new fixture confirmed to bind under a stubbed `DerivedRuleName`.

The `RuleNames` nit is yours to take or leave; it is correct as written.

---

PR #210, opened 2026-08-12
End an indefinite-length tuple where its arity says, closes #199

Rafael-SOWNet:
Closes #199, as you specified it. **1717 tests**, +3 here, green on net10.0; all four TFMs build.

The trailing check is now the mirror of the leading ones at each of the seven arities, so an indefinite-length tuple ends where its arity says it does and the break is consumed.

## Two notes on the implementation

**`reader.ReadEndArray()` does not exist.** The suggested fix names it, and `CborReader` has no such method — `ConsumeBreak` is what `ReadChunksArray` and the skip paths use, and it was `private`. It is now `internal`, which is what `IsBreak` already is and what this converter already depends on. If you would rather have a public `ReadEndArray()` for the symmetry with `CborWriter.WriteEndArray(size)` — it is the method a hand-written array-shaped converter needs, and its absence is arguably why this defect was writable in the first place — say so and I will add it here instead.

**`Tuple8Converter` gets the check too, and nothing can reach it.** #198 keeps every arity past seven from building a converter at all. Same call as #189, which put the major-type check there for the same reason: the code is right for when the arity becomes reachable, rather than needing to be revisited then. It does mean that one of the seven is unasserted, which is what #198's own note about #189's coverage records.

## Tests

The three cases you named:

* **Too many items refused at every arity**, both length forms — the definite-length half already held from `size != N`, and is asserted alongside so the two stay together.
* **A well-formed indefinite-length tuple still reads at every arity.** This is the guard against the easy way to make the first test pass.
* **Nothing left behind**, in the two places the stray break did its damage: `{"a": [_ 1, 2], "b": 7}` now reads as `(1, 2)` and `7`, and `[[_ 1, 2], [_ 3, 4]]` as two tuples. Both were `CborException`s before, at `$` and `$[1]`.

Checked by reverting the converter and re-running: the first and third fail, the second passes either way, as its remark says it should.

## Scope

**This does not close #198.** It is the other half of what you found reviewing #189, and the arity defect needs a decision this PR does not take — the wire format for arity 8 and up. `Tuple8Converter` would have to write its `Rest` field's items *inline* rather than as a nested array, since a nine-element tuple should be `[1, …, 9]` and not `[1, …, 7, [8, 9]]`, and that means every arity growing an "write my items into the array you already opened" entry point plus support for the one-element `Rest` the provider currently refuses outright. Nothing works there today, so there is no wire compatibility to preserve and the flat form is free to be chosen — but it is a redesign of that file rather than a fix, and refusing arity above seven with a plain `CborException` is a defensible alternative that costs five lines. I would rather you picked. Happy to build either.

---

### DIFF
--- a/src/Dahomey.Cbor.Tests/TupleTests.cs
+++ b/src/Dahomey.Cbor.Tests/TupleTests.cs
@@ -1,5 +1,6 @@
 using Dahomey.Cbor.Attributes;
 using Dahomey.Cbor.Tests.Extensions;
+using System.Collections.Generic;
 using Xunit;
 
 namespace Dahomey.Cbor.Tests
@@ -153,6 +154,63 @@ namespace Dahomey.Cbor.Tests
 
             Assert.Contains("Expected major type Array", exception.Message);
             Assert.Equal("$.a", exception.Path);
+        }
+
+        /// <summary>
+        /// An indefinite-length tuple ends where its arity says it does, at every arity and in both
+        /// length forms. The definite-length half already held, from <c>size != N</c>; the
+        /// indefinite-length half had no check after the last item at all, so the extra items were
+        /// silently dropped.
+        /// </summary>
+        [Fact]
+        public void ATupleWithTooManyItemsIsRefusedAtEveryArity()
+        {
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int)>("9F010203FF".HexToBytes()));
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int, int)>("9F01020304FF".HexToBytes()));
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int, int, int)>("9F0102030405FF".HexToBytes()));
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int, int, int, int)>("9F010203040506FF".HexToBytes()));
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int, int, int, int, int)>("9F01020304050607FF".HexToBytes()));
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int, int, int, int, int, int)>("9F0102030405060708FF".HexToBytes()));
+
+            // The definite-length form of the same document, which the arity check already refused.
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int)>("83010203".HexToBytes()));
+            Assert.Throws<CborException>(() => Cbor.Deserialize<(int, int, int)>("8401020304".HexToBytes()));
+        }
+
+        /// <summary>
+        /// The other side of that check: a well-formed indefinite-length tuple still reads, at every
+        /// arity. Refusing one would be the easy way to make the test above pass.
+        /// </summary>
+        [Fact]
+        public void AnIndefiniteLengthTupleIsStillReadAtEveryArity()
+        {
+            Assert.Equal((1, 2), Helper.Read<(int, int)>("9F0102FF"));
+            Assert.Equal((1, 2, 3), Helper.Read<(int, int, int)>("9F010203FF"));
+            Assert.Equal((1, 2, 3, 4), Helper.Read<(int, int, int, int)>("9F01020304FF"));
+            Assert.Equal((1, 2, 3, 4, 5), Helper.Read<(int, int, int, int, int)>("9F0102030405FF"));
+            Assert.Equal((1, 2, 3, 4, 5, 6), Helper.Read<(int, int, int, int, int, int)>("9F010203040506FF"));
+            Assert.Equal((1, 2, 3, 4, 5, 6, 7), Helper.Read<(int, int, int, int, int, int, int)>("9F01020304050607FF"));
+        }
+
+        /// <summary>
+        /// Where the unconsumed break did its damage: the closing <c>FF</c> was left in the buffer, so
+        /// whatever followed the tuple read it instead. An indefinite-length tuple was therefore only
+        /// usable as a whole document — as a map member the break was read as the next key, and as an
+        /// array element as the next element's head.
+        /// </summary>
+        [Fact]
+        public void AnIndefiniteLengthTupleLeavesNothingBehind()
+        {
+            // {"a": [_ 1, 2], "b": 7}
+            TupleMember member = Cbor.Deserialize<TupleMember>("A261619F0102FF616207".HexToBytes());
+
+            Assert.Equal((1, 2), member.A);
+            Assert.Equal(7, member.B);
+
+            // [[_ 1, 2], [_ 3, 4]]
+            List<(int, int)> list = Cbor.Deserialize<List<(int, int)>>("829F0102FF9F0304FF".HexToBytes());
+
+            Assert.Equal(new[] { (1, 2), (3, 4) }, list);
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


mcatanzariti (commented):
The fix is right and the tests hold. Verified rather than read: merged onto `master` at 4bfc252 (four commits past your base — `CborReader.cs` has grown by 356 lines since, and it still merges clean), all four TFMs build, and `CDDL_REQUIRED=1 dotnet test` is green on `net10.0` and `net8.0` — 1889 tests. Reverting `TupleConverter.cs` alone fails `ATupleWithTooManyItemsIsRefusedAtEveryArity` and `AnIndefiniteLengthTupleLeavesNothingBehind`, and leaves `AnIndefiniteLengthTupleIsStillReadAtEveryArity` passing, exactly as your remark says it should.

One decision below is mine to make and I make it; the rest is one missing test, one factoring, and two notes.

## The argument this PR does not make, and should

The defect broke a round-trip through this library's own writer. With `ArrayLengthMode.IndefiniteLength`:

```csharp
var options = new CborOptions { ArrayLengthMode = LengthMode.IndefiniteLength };
string hex = Helper.Write(new List<(int, int)> { (1, 2), (3, 4) }, options);  // 9F9F0102FF9F0304FFFF
Cbor.Deserialize<List<(int, int)>>(hex.HexToBytes());                         // CborException before, correct after
```

Nothing hand-written is involved: the library could not read back what it had just written, for any tuple in any container, as soon as that option was set. That is a better statement of the severity than the hex repros in #199, and it is worth a fourth test — `AnIndefiniteLengthTupleLeavesNothingBehind` covers the same mechanism but only from documents typed by hand.

Six other shapes I probed were broken before this PR and are correct after it: an indefinite-length *outer* array of tuples (`9F 9F0102FF 9F0304FF FF`), a nested tuple (`((int,int),int)`), an indefinite-length map holding a tuple, a `(int, int)?` member followed by another member, a tuple whose item is an indefinite-length array, and a truncated `9F0102`, which used to yield `(1, 2)` and now raises. That last one is a behaviour change for documents that were malformed all along — intended, and worth a line in the release notes.

## `ConsumeBreak` internal: keep it

You are right that `ReadEndArray()` does not exist; I wrote the name in #199 from `CborWriter`'s side of the pair. `internal` matches `IsBreak`, which this converter already depends on, so take it as it stands here.

One fact against adding the public method in this PR: `ReadArray` does not call `ConsumeBreak` either — it ends with `_state = CborReaderState.Start;` inline (`CborReader.cs:1188`). So there are already two idioms for the same act, and a third entry point would have to absorb that one to be worth it. That is a separate change to `CborReader`, not part of ending a tuple; I would rather have it as its own PR, with `ReadArray` converted in the same pass.

## Factor the check instead of writing it seven times

This is the one thing I would like changed here. The end-of-array check is the seventh copy of something that has to be right in seven places, and it sits next to 35 copies of the per-item check and 7 of the head, message strings included. The missing end was writable precisely because there was no single place to put it — adding a seventh copy leaves the next check in the same position.

The file already has `TupleItemReader` as the home for what the arities share. Widening it to cover the whole array — head, item, end — takes `TupleConverter.cs` from 704 lines to 440, and each arity becomes the list of items it is:

```csharp
public override (T1, T2) Read(ref CborReader reader)
{
    int size = TupleReader.ReadBeginArray(ref reader, 2);

    T1 item1 = TupleReader.ReadItem(ref reader, _item1Converter, size, 2, 0);
    T2 item2 = TupleReader.ReadItem(ref reader, _item2Converter, size, 2, 1);

    TupleReader.ReadEndArray(ref reader, size, 2);

    return (item1, item2);
}
```

<details>
<summary>The helper, with the three checks and the comment that was pasted seven times</summary>

```csharp
/// <summary>
/// The array-shaped reading that every arity of tuple does identically: the head, one item at a
/// position, and the end.
/// </summary>
/// <remarks>
/// Written once here rather than in each of the seven converters, which otherwise differ only in
/// how many items they read - a check that has to be right in seven places is a check that gets
/// added to six of them, which is how the missing end was writable in the first place.
/// </remarks>
internal static class TupleReader
{
    /// <summary>
    /// Reads the array head and returns its size, or -1 for an indefinite-length array.
    /// </summary>
    /// <remarks>
    /// ReadBeginArray rather than going straight to ReadSize. [the existing comment, once]
    /// </remarks>
    public static int ReadBeginArray(ref CborReader reader, int arity)
    {
        reader.ReadBeginArray();

        int size = reader.ReadSize();

        if (size != -1 && size != arity)
        {
            throw new CborException($"Expected CBOR Array of size {arity}");
        }

        return size;
    }

    /// <summary>
    /// Reads one item of a tuple, naming its position in <see cref="CborException.Path"/> if it fails.
    /// </summary>
    /// <remarks>
    /// A tuple has no member names to report, so the position is all there is to say - and it is what
    /// a caller needs, because the items of a tuple are otherwise indistinguishable in a message.
    /// <para>
    /// An indefinite-length array is checked for its break first: an array that ends before the arity
    /// does is too short, rather than an item to read. That exception carries no position, since the
    /// item it would name is the one that is not there.
    /// </para>
    /// </remarks>
    public static T ReadItem<T>(ref CborReader reader, ICborConverter<T> converter, int size, int arity, int index)
    {
        if (size == -1 && reader.IsBreak())
        {
            throw new CborException($"Expected CBOR Array of size {arity}");
        }

        try
        {
            return converter.Read(ref reader);
        }
        catch (CborException exception)
        {
            exception.PrependPathIndex(index);
            throw;
        }
    }

    /// <summary>
    /// Ends an indefinite-length tuple where its arity says it does, and consumes the break.
    /// </summary>
    /// <remarks>
    /// The mirror of the check <see cref="ReadItem"/> makes before each item: anything other than the
    /// break here is an item past the arity, which the definite-length form already refuses from its
    /// size. Consuming it is what leaves the reader positioned after the tuple, so an indefinite-length
    /// tuple is readable as a map member or an array element and not only as a whole document. A
    /// definite-length array has nothing to end.
    /// </remarks>
    public static void ReadEndArray(ref CborReader reader, int size, int arity)
    {
        if (size != -1)
        {
            return;
        }

        if (!reader.IsBreak())
        {
            throw new CborException($"Expected CBOR Array of size {arity}");
        }

        reader.ConsumeBreak();
    }
}
```

</details>

Built and tested on top of this PR, on `claude/analyse-pr-210-9mtp14` in this repository (your commit merged onto current `master`, then the factoring as one commit — cherry-pick it or rewrite it, whichever you prefer). `126 insertions(+), 390 deletions(-)`. No behaviour change: the messages are the same strings — interpolation only runs on the throwing path — the too-short exception still carries no position, a failing item still carries `$[1]`, and the suite is green on both TFMs. It does drag the head and the per-item check into this PR's diff, which is beyond what #199 asks for; I think that is the right trade for this file, but if you would rather keep the diff to the end of the array, factor `ReadEndArray` alone and leave the other two.

While you are in there: `9F01FF` — the too-short case, which the leading checks already refuse — is still asserted nowhere. It is one line in `ATupleWithTooManyItemsIsRefusedAtEveryArity`, and it pins the check your new one is the mirror of.

## The two notes

**On #198.** Right not to close it, and the wire-format analysis is right: a nine-element tuple should be `[1, …, 9]`, which means an entry point for writing items into an already-open array and support for the one-element `Rest`. Confirming your reading of the reachability, and adding one thing to it — today `(int, int, int, int, int, int, int, int)` throws `TargetInvocationException`, not `CborException`, because the provider's `case 8` builds `Tuple8Converter` whose `TRest` converter lookup hits the `default` arm for a 1-tuple. So the cheap option is not the status quo plus five lines: it is a real improvement over an unwrapped reflection exception. I will decide the format in #198 rather than here.

**Follow-up, not for this PR.** `TupleConverter` still does not go through `EnterNestedItem`, so `MaxDepth` is not counted across tuple frames the way it is for `ReadArray` and `ReadMap`. Harmless in practice — tuple nesting is fixed by the C# type and not by the message, so a payload cannot drive it — but it is the remaining half of "the one array-shaped reader that does not go through `ReadArray`" that #199 and #187 both circle.

---

PR #211, opened 2026-08-12
Report two members mapped to one CBOR key, closes #194

### LINKED ISSUE #194: The source generator has no diagnostic for two members mapped to the same CBOR name
#177 made a type that maps two members to one CBOR name refuse to build its mapping, so the reflection path now reports the collision the first time the type is used. The generator does not: a `[CborSerializable]` type with two members under one name compiles clean, and the failure appears later, when the context is constructed.

```csharp
[CborSerializable]
public partial class Aliased
{
    [CborProperty("X")] public int First { get; set; }
    [CborProperty("X")] public int Second { get; set; }
}
```

The generator is the other place a mapping is made, and it is the place that can say so earliest — at the keystroke, against the second attribute, rather than at the first `new MyContext(options)`. `Diagnostics.cs` already carries ten descriptors for exactly this kind of thing, `MissingMemberIndex` among them, so the shape of the answer is settled; what is missing is the rule and its descriptor.

This is a diagnostic rather than a correctness gap — the runtime check catches the same types, only later, and the generated converter is not wrong about anything the reflection path is right about. That argues for treating it as a follow-up rather than as part of #186, which is where it was noticed.

Worth deciding while writing it: whether the generator can see every route the runtime check covers. The attribute is visible in the syntax, and so is a naming convention applied through `[CborNamingConvention]` if the convention type can be resolved and run at generation time. A member hiding a base member is visible too, though the base may be in another assembly. Anything the generator cannot see stays with the runtime check, which is not a problem as long as the diagnostic does not claim to be exhaustive.

Rafael-SOWNet:
Closes #194. **1717 library tests plus 39 generator tests**, +3 and +4 here, green on net10.0; all four TFMs build.

`CBOR1013` for two members under one CBOR name, against the second declaration rather than at the first `new MyContext(options)`. `CBOR1014` for the same failure where the wire key is the index rather than the name — `IntKeyMap` and `Array` — which is the same split `ObjectConverter`'s read lookup makes. Leaving that half out would have caught one of the two things #186 refuses and invited a follow-up issue.

Errors rather than warnings, because since #186 such a type cannot serialize at all: there is no working program this turns away.

## Your open question, measured

> whether the generator can see every route the runtime check covers

It cannot, and one of the routes you listed is closed to it for a reason worth knowing: **a `[CborNamingConvention]` on a type is already `CBOR1007`**, since the generator does not reproduce type-level conventions at all. So a convention folding two names into one on that route cannot reach this check, and the run-time refusal stays the only thing covering it. The *context-level* `NamingConvention` is reproduced, and folding two member names into one under that is reported — which is the case worth having, because nothing there is a duplicate as declared.

A member hiding a base member is visible and is reported. What stays with the run-time check: a hidden member whose base is in another assembly with no source, and anything a convention type the generator cannot run decides. The descriptor's remark says it is not exhaustive rather than implying it is.

## An overridden property was already broken, and this had to fix it first

The false-positive test I wrote for overrides failed, and the reason is not this diagnostic.

`EnumerateMembers` walks the type and each of its bases, so an override is declared twice in what the generator sees. `Type.GetProperties` collapses an override onto its base, so the reflection path has always seen one. Two mappings under one name wrote the key twice — and since #186 validates the mapping, **a generated context over any type with an overridden property now throws while being built.** Measured on `e2689ef` with these tests and the generator reverted:

```
CborException: class/struct GeneratedOverrideHolder maps several fields/properties
  to the member name 'Id'
    at ObjectMapping.ValidateMemberNamesAndindexes
```

Before #186 it silently wrote the key twice instead, so the defect is older than yesterday; #186 turned it from wrong bytes into a hard failure. Either way `virtual`/`override` is ordinary enough in a serialized model that I would not leave it, and CBOR1013 would otherwise fire for a type that has one member and not two.

So the base declaration is dropped once an override for it has been seen — the walk runs most-derived first, so the override always comes first. `new`-hiding is deliberately *not* collapsed: `GetProperties` returns both, the reflection path refuses the mapping, and CBOR1013 says the same thing earlier.

`GeneratedOverrideTests` pins it, asserting the map has two members rather than three and that the bytes match the reflection path's; the type is enrolled in `GeneratedCorpusTests` too, so it is covered by the comparison that exists for exactly this class of divergence.

If you would rather have that half as its own PR against its own issue, say so and I will split it — I kept it here because the diagnostic is a false positive without it, so the two cannot land separately in this order.

---

### DIFF
--- a/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
+++ b/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
@@ -332,6 +332,67 @@ public partial class Context : CborSerializerContext { }
         }
 
         /// <summary>
+        /// The same, under <c>Array</c>. The descriptor's summary claims both index-keyed formats and
+        /// the branch is reached through <c>ObjectFormat != "StringKeyMap"</c>, so <c>IntKeyMap</c>
+        /// alone leaves the other half of the claim resting on reading the condition.
+        /// </summary>
+        [Fact]
+        public void TwoMembersUnderOneCborIndexAreReportedUnderArrayFormatToo()
+        {
+            AssertReports("CBOR1014", @"
+[CborObjectFormat(CborObjectFormat.Array)]
+public class Indexed
+{
+    [CborProperty(1)] public int First { get; set; }
+    [CborProperty(1)] public int Second { get; set; }
+}
+
+[CborSerializable(typeof(Indexed))]
+public partial class Context : CborSerializerContext { }
+");
+        }
+
+        /// <summary>
+        /// A member hiding a base member with <c>new</c> is two members, and stays reported. That is why
+        /// the collapse filters on <c>IsOverride</c> rather than on the name: <c>Type.GetProperties</c>
+        /// returns both declarations, so the reflection path sees two members under one name and refuses
+        /// the mapping, and the generator gives the same answer earlier. Widening the filter to names
+        /// would turn this into a silent drop of whichever declaration came second.
+        /// </summary>
+        [Fact]
+        public void AMemberHidingABaseMemberWithNewIsReported()
+        {
+            AssertReports("CBOR1013", @"
+public class Shape { public int Id { get; set; } }
+
+public class Circle : Shape { public new int Id { get; set; } }
+
+[CborSerializable(typeof(Circle))]
+public partial class Context : CborSerializerContext { }
+");
+        }
+
+        /// <summary>
+        /// The collapse through a <em>constructed</em> generic base, which is the fragile half of the
+        /// mechanism: it leans on <c>OverriddenProperty</c> yielding the member of <c>Holder&lt;int&gt;</c>
+        /// and on <c>SymbolEqualityComparer.Default</c> matching that against what <c>GetMembers()</c>
+        /// returns for the same constructed type. Without it this is CBOR1013, and every generated
+        /// context over a type overriding a generic base member fails to build.
+        /// </summary>
+        [Fact]
+        public void AnOverrideThroughAGenericBaseIsNotReported()
+        {
+            AssertClean(@"
+public class Holder<T> { public virtual T Value { get; set; } }
+
+public class IntHolder : Holder<int> { public override int Value { get; set; } }
+
+[CborSerializable(typeof(IntHolder))]
+public partial class Context : CborSerializerContext { }
+");
+        }
+
+        /// <summary>
         /// An abstract base is never instantiated — its subtypes are — so it must not be reported.
         /// </summary>
         [Fact]
--- a/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
+++ b/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
@@ -247,6 +247,91 @@ public partial class Context : CborSerializerContext { }
         }
 
         /// <summary>
+        /// Two members under one CBOR name. #186 made the reflection path refuse the mapping when it is
+        /// built, so such a type cannot serialize at all; the generator is where it can be said first,
+        /// against the second declaration rather than at the first construction of the context.
+        /// </summary>
+        [Fact]
+        public void TwoMembersUnderOneCborNameAreReported()
+        {
+            AssertReports("CBOR1013", @"
+public class Aliased
+{
+    [CborProperty(""X"")] public int First { get; set; }
+    [CborProperty(""X"")] public int Second { get; set; }
+}
+
+[CborSerializable(typeof(Aliased))]
+public partial class Context : CborSerializerContext { }
+");
+        }
+
+        /// <summary>
+        /// The other route the generator can see, and the one worth catching early: a naming policy
+        /// folding two member names into one, where nothing is a duplicate as declared.
+        /// </summary>
+        /// <remarks>
+        /// The policy has to be the context-level one. A <c>[CborNamingConvention]</c> on the type is
+        /// already CBOR1007 — the generator does not reproduce it at all — so that route cannot reach
+        /// this check, and the run-time refusal remains the only thing covering it.
+        /// </remarks>
+        [Fact]
+        public void ANamingPolicyFoldingTwoNamesIntoOneIsReported()
+        {
+            AssertReports("CBOR1013", @"
+public class Folded
+{
+    public int Value { get; set; }
+    public int value { get; set; }
+}
+
+[CborSerializable(typeof(Folded))]
+[CborSourceGenerationOptions(
+    NamingConvention = typeof(Dahomey.Cbor.Serialization.Conventions.CamelCaseNamingConvention))]
+public partial class Context : CborSerializerContext { }
+");
+        }
+
+        /// <summary>
+        /// The same failure where the wire key is the index rather than the name, which is what
+        /// <c>IntKeyMap</c> and <c>Array</c> are.
+        /// </summary>
+        [Fact]
+        public void TwoMembersUnderOneCborIndexAreReported()
+        {
+            AssertReports("CBOR1014", @"
+[CborObjectFormat(CborObjectFormat.IntKeyMap)]
+public class Indexed
+{
+    [CborProperty(1)] public int First { get; set; }
+    [CborProperty(1)] public int Second { get; set; }
+}
+
+[CborSerializable(typeof(Indexed))]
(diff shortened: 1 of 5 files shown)


src/Dahomey.Cbor.Generator/TypeCollector.cs
```diff
@@ -290,8 +332,25 @@ private static IEnumerable<ISymbol> AllMembers(ITypeSymbol type)
...
+        /// <para>
+        /// A member hiding a base member with <c>new</c> is deliberately not collapsed.
+        /// <c>GetProperties</c> returns both of those, so the reflection path sees two members under one
+        /// name and refuses the mapping; the generator reports <c>CBOR1013</c> for the same type, which
+        /// is the same answer given earlier.
+        /// </para>
+        /// </remarks>
         private IEnumerable<ISymbol> EnumerateMembers(ITypeSymbol type)
         {
+            HashSet<ISymbol> overriddenByADerivedMember =
```
mcatanzariti, line 428: Collapsing on the most-derived declaration matches `Type.GetProperties` on member *identity*, but not on where the attributes come from. The reflection path reads them with inheritance — `MemberInfo.GetCustomAttribute<CborPropertyAttribute>()` in `MemberMapping` and `IsDefined(typeof(CborIgnoreAttribute))` in `DefaultObjectMappingConvention` both go through the `inherit: true` overloads — while this walk now keeps only the override symbol, and `ISymbol.GetAttributes()` does not include the base declaration's.

Measured on this branch, against the reflection path:

```csharp
class Base    { [CborProperty("id")] public virtual int Id { get; set; } }
class Derived : Base { public override int Id { get; set; } public string Name { get; set; } }
```

```
reflection : A2 62 6964 07 …   → key "id"
generated  : A2 62 4964 07 …   → key "Id"     ← no diagnostic
```

`[CborIgnore]` on the virtual base declaration goes the same way: reflection writes a map of one, the generated context writes a map of two, silently.

The `[CborIgnore]` half is pre-existing and unchanged by this PR. The `[CborProperty]` half is not: on master both declarations were emitted, so the type wrote two keys and, since #186, threw while the context was built. With the collapse it writes one key, under the wrong name, and says nothing — a loud failure turned into a silent divergence, which is the class of thing the generator's contract makes a build error and `GeneratedCorpusTests` exists to catch.

Two ways out, either is fine: carry the attributes of the collapsed base declaration when the override does not declare its own, or report a diagnostic when the declaration being dropped carried a `[CborProperty]` or `[CborIgnore]`. Whichever, a type in `GeneratedCorpusTests` would pin it against the reflection path.

---

src/Dahomey.Cbor.Generator/Diagnostics.cs
```diff
@@ -146,6 +146,41 @@ internal static class Diagnostics
         /// that cannot be instantiated at all -- an abstract class or an interface -- since a concrete
         /// class with no subtypes is simply described by its own rule.
         /// </summary>
+        /// <summary>
```
mcatanzariti, line 157: The new block was inserted below CBOR1012's existing doc comment rather than after its descriptor, so the comment moved owners: `DuplicateMemberName` now carries **two `<summary>` elements** — the polymorphic-schema one at l. 143-148 and its own at l. 149-154 — and `IncompletePolymorphicSchema` (l. 184) is left with no documentation at all.

No build warning, since the generator project does not emit a documentation file, but the first summary on CBOR1013 describes a different diagnostic.

Moving the CBOR1012 summary back down to its descriptor fixes it. While there: CBOR1013/1014 are declared ahead of CBOR1012, which is the only place in this file where the ids run out of order.

---

src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
```diff
@@ -246,6 +246,91 @@ public partial class Context : CborSerializerContext { }
...
+        }
+
+        /// <summary>
+        /// The same failure where the wire key is the index rather than the name, which is what
+        /// <c>IntKeyMap</c> and <c>Array</c> are.
+        /// </summary>
+        [Fact]
+        public void TwoMembersUnderOneCborIndexAreReported()
+        {
+            AssertReports("CBOR1014", @"
```
mcatanzariti, line 302: Three behaviours this PR relies on are asserted in the description but not pinned by a test. I ran each against this branch — all three already behave as claimed, so these are cheap to add and they are what stops the next change from moving them quietly:

1. **`new`-hiding still reports CBOR1013.** The PR argues the collapse is deliberately not applied to hidden members because the reflection path refuses that mapping. Measured: `CBOR1013: 'Derived' maps both 'Id' and 'Id' …`. That distinction is the whole reason `EnumerateMembers` filters on `IsOverride` rather than on the name, and nothing currently fails if someone widens it.

2. **CBOR1014 fires for `Array`, not only `IntKeyMap`.** `TwoMembersUnderOneCborIndexAreReported` covers `IntKeyMap`; the descriptor's summary claims both formats, and the branch is reached through `model.ObjectFormat != "StringKeyMap"`. Measured on `[CborObjectFormat(CborObjectFormat.Array)]`: `CBOR1014: 'Indexed' maps both 'First' and …`.

3. **An override through a generic base is collapsed too.** `GeneratedOverrideTests` uses a non-generic base, while the collapse leans on `OverriddenProperty` returning the member of the *constructed* base and on `SymbolEqualityComparer.Default` matching it against what `GetMembers()` yields for `Base<int>`. It does work — but it is the fragile half of the mechanism and it is the untested one. On master `class Derived : Base<int>` with an `override` throws `class/struct ProbeGenericDerived maps several fields/properties to the member name 'Value'` from `ProbeContext.Configure`, which is the same defect this PR fixes, reached by a route no test here takes.

---

mcatanzariti (commented):
Reviewed against a local build and run rather than by reading: SDK 10 plus the `cddl` gem, all four TFMs building, and the suite run twice — on `08bd1a7` (**1782 + 39 green**, matching the rebase comment) and on the merge with today's `master` at `4bfc252`, i.e. after #207, #209 and #212 (**1889 + 39 green**). The only conflict against current `master` is the `GeneratedCorpusTests.Sample` one already described, and nothing in #207's mapping changes collides with this.

The central claim holds, and I reproduced it independently with a type this PR does not contain — a *generic* base whose property is overridden. On today's `master`, building the generated context throws:

```
CborException : class/struct ProbeGenericDerived maps several fields/properties to the member name 'Value'
  at ObjectMapping.ValidateMemberNamesAndindexes
  at ProbeContext.Configure          ← generated code, while the context is being built
```

So a generated context over any type with an overridden property is broken today, the override fix is not optional, and CBOR1013 would indeed be a false positive without it. I would not split the PR: in the other order the fix needs an issue of its own for a defect this PR already documents with its repro.

The name/index split also matches the runtime exactly — the same `switch (ObjectFormat)` as `ValidateMemberNamesAndindexes`, `StringComparer.Ordinal` included — and I confirmed the deliberate negative case, `IntKeyMap` with duplicate names under distinct indexes, reports nothing. `cborIndex!.Value` is safe: the `MissingMemberIndex` check above it already `continue`s on null for every non-`StringKeyMap` format.

Three things before merge, inline:

1. **A silent divergence the collapse leaves open** — attributes on the overridden base declaration are not carried, so `[CborProperty("id")]` on a virtual base makes the generated context write `Id` where the reflection path writes `id`, with no diagnostic. This is the one worth acting on: for `[CborProperty]` the PR turns a build-time throw into wrong bytes.
2. **CBOR1012's doc comment now sits on CBOR1013**, which ends up with two `<summary>` elements while `IncompletePolymorphicSchema` has none.
3. **Three missing tests** for behaviours the description asserts and nothing pins: `new`-hiding, CBOR1014 under `Array`, and an override through a generic base.

Minor, take or leave: the `if / else if / else` in `CollectObject` mixes the format choice with the duplicate check and reads harder than two symmetric nested `if`s would; and if someone sets `dotnet_diagnostic.CBOR1013.severity = none`, the `continue` drops the member from the model silently, which assumes the error always stops the build.

---

mcatanzariti:
Merging this as it stands. All three review points are closed and verified, and the two minor notes with them.

The one thing still open is the wording of the comment at `TypeCollector.cs:252-259`, not the code it sits above — the behaviour it describes stays. I am taking that to a follow-up PR rather than holding this one for it, so there is nothing left to push here.

Thanks for the measurements throughout; the override defect this uncovered was worth more than the diagnostic it started as.

---

PR #214, opened 2026-08-13
Read and write a tuple of any arity as one flat array, closes #198

Rafael-SOWNet:
Closes #198, taking the flat encoding from the [investigation](https://github.com/dahomey-technologies/Dahomey.Cbor/issues/198#issuecomment-5280922038), and makes tuples work under Native AOT while it is there. **1900 library tests plus 35 generator tests**, green on net10.0; all four TFMs build.

Two commits, each standing on its own: the runtime fix, then the generator.

## The wire form

```
(1, "two")                    ->  82 01 63 74776F
(1, …, 9)                     ->  89 01 02 03 04 05 06 07 08 09
fifteen elements              ->  8F 01 … 0F
```

C# represents a tuple past seven elements as seven fields plus a `Rest`, and nests again every seven after that. That is a detail of the language, not something the format knows, so the overflow's converter contributes its items to the array the outer one opened rather than opening one of its own. A nine-element tuple is nine items.

## Why the eight converters cover every arity

This is the part that made it cheap. `Tuple8Converter<T1, …, T7, TRest>` delegates to `TRest`'s own items, so it handles arity 8 through 14 with a flat `Rest`, then 15 through 21 by recursing into itself, and so on without limit. **The provider's `default:` is now unreachable for a real tuple** — C# never gives one more than eight fields.

That is variadic in the only sense C# offers, and it is how the BCL states the same thing: `ValueTuple` has seven flat arities plus a recursive eighth, and so now do the converters. The two genuinely variadic alternatives are both worse here — `ITuple` boxes every element and cannot construct one back, and reflection over the fields is what a generated context exists to avoid.

`Tuple1Converter` is new because an eight-element tuple's `Rest` is exactly a one-element tuple. C# has no literal for one; it is an array of one item like any other tuple.

## It is a net deletion

The array around the items is now stated once, in `TupleReader` and `TupleWriter`: `ReadBeginArray`, the size check, the break checks between items, the trailing break, and the length mode. Each arity keeps only its items. **`TupleConverter.cs` is 80 lines shorter than when it covered seven arities**, and `Tuple2Converter` went from 58 lines to 37.

The eighth element was never wrong in the converter, incidentally — `Tuple8Converter` already declared `CborConverterBase<(T1, …, T8)>` and read eight items. Only what the provider handed it was.

## Native AOT

A tuple member on a `[CborSerializable]` type was `CBOR1002`: the generator saw an unrecognised generic type and refused to build. So tuples were unavailable to exactly the applications the generated path exists for, since the reflection provider resolves their converters through `Type.MakeGenericType`.

The generator now classifies a `ValueTuple` as its own kind and writes the instantiation out — `new Tuple2Converter<int, string>`, and for a longer tuple `new Tuple8Converter<…, ValueTuple<int, int>>` together with the `Rest`'s own converter, which the dependency ordering registers first because `Tuple8Converter` asks the registry for it while being constructed.

Detection is by metadata name rather than `IsTupleType`, which is **false** for the one-element `ValueTuple<T>` — C# has no tuple syntax for a single element, and that arity is exactly what an eight-element tuple's `Rest` is.

**Proven on a real native binary**, not on the analyzers: a probe with `PublishAot` inside the csproj and gcc as the linker, over a context carrying arity 2, 9 and 15 plus a nullable tuple.

```
written : A4645061697282016374776F644E696E6589010203040506070809674669667465656E8F01…0F684F7074696F6E616C820765736576656E
expected: A4645061697282016374776F644E696E6589010203040506070809674669667465656E8F01…0F684F7074696F6E616C820765736576656E
ROUND TRIP OK
```

The expected bytes are the reflection path's, captured on CoreCLR, so the native binary is held to byte identity rather than to merely producing something. The probe exercises only the generated path, since a plain `CborOptions` over the same model is the failure the context exists to prevent.

## What still cannot be AOT, and why

Documented in the README rather than left implicit. **Without a context, a tuple needs the reflection provider, which Native AOT cannot serve** — `TupleConverterProvider` reads the fields and calls `MakeGenericType`, and the AOT analyzer says so (`IL2070` on `GetFields`, `IL3050` on `MakeGenericType`). That is the same requirement as for every other generic converter here — collections, dictionaries, nullables — and not specific to tuples, so the README says it that way. `System.Tuple<…>`, the class, is not a tuple to this library at all and is unchanged.

## CDDL

A tuple renders as a fixed heterogeneous array with the `Rest` chain flattened — `[int, tstr]` — because that is what the writer emits. Not `[* X]`, which would say any length of one type. Validated with the gem against a real instance.

## Interaction with #210

This supersedes it: the trailing-break check #199 asked for now lives in `TupleReader.Read` instead of in each of the seven arities, and `ConsumeBreak` becomes `internal` here too. Happy to close #210, or to rebase this on top of it if you would rather have that one first — say which.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ High-performance [CBOR](https://cbor.io/) serialization framework for .Net (C#)
 * Support for structs
 * Support for RFC 8746 typed arrays, opt-in per direction (CborOptions.TypedArrayMode)
 * Support for RFC 8949 §3.4.3 bignums (tags 2 and 3) as System.Numerics.BigInteger
+* Support for tuples of any arity, as one flat array, on the reflection and source-generated paths
 * Reads RFC 8949 §3.4.4 decimal fractions (tag 4) into decimal, and writes them on request
   (CborOptions.DecimalFormat)
 * Support for RFC 8949 §3.4.4 decimal fractions (tag 4) and bigfloats (tag 5) as their own types,
@@ -390,6 +391,47 @@ Equality is structural over the pair as encoded, so `10e0` and `1e1` are the sam
 equal, and neither is normalised on the way out — a document round-trips byte for byte. Two encodings of
 one number are therefore two distinct dictionary keys, and under `CborOptions.Deterministic` they sort as
 the different byte strings they are.
+
+### Tuples
+
+A `ValueTuple` is one CBOR array holding its elements in order, at any arity:
+
+```csharp
+// (1, "two")                        -> 82 01 63 74776F
+// (1, 2, 3, 4, 5, 6, 7, 8, 9)       -> 89 01 02 03 04 05 06 07 08 09
+// fifteen elements                  -> 8F 01 02 ... 0F
+```
+
+**Past seven elements the array stays flat.** C# represents such a tuple as seven fields plus a `Rest`
+holding the overflow, and nests again every seven after that, but that is a detail of the language: a
+nine-element tuple is nine items, not seven and a nested pair. A one-element `ValueTuple<T>` — which C#
+has no literal for — is an array of one.
+
+Reading is exact about arity: an array of the wrong length is a `CborException` naming the arity
+expected, in both the definite and indefinite-length forms, and a failure inside an element names its
+flattened position, so `$[8]` rather than a path through a `Rest` the document knows nothing about.
+
+`System.Tuple<…>` — the class, not the struct — is not a tuple to this library and is treated as an
+ordinary object.
+
+#### Under Native AOT
+
+Tuples work in a source-generated context, at every arity:
+
+```csharp
+[CborSerializable(typeof(Reading))]
+public partial class MyContext : CborSerializerContext { }
+```
+
+The generated context names each converter instantiation in source — `Tuple2Converter<int, string>`, and
+for a longer tuple `Tuple8Converter<…, ValueTuple<int, int>>` together with the `Rest`'s own converter —
+so nothing is constructed reflectively and the bytes are identical to the reflection path's.
+
+**Without a context, a tuple needs the reflection path, which Native AOT cannot serve.** The provider
+resolves a tuple's converter through `Type.MakeGenericType`, as it does for every generic converter —
+collections, dictionaries, nullables — so a trimmed or AOT-published application must declare the types
+it serializes on a context. That is the same requirement as for any other generic type here, not
+something specific to tuples.
 
 ### Custom converters
 
--- a/src/Dahomey.Cbor.Generator/CddlTypeReference.cs
+++ b/src/Dahomey.Cbor.Generator/CddlTypeReference.cs
@@ -139,6 +139,29 @@ namespace Dahomey.Cbor.Generator
                     break;
                 }
 
+                // A fixed, heterogeneous array: one entry per element, in order, with the Rest chain
+                // flattened because that is what the writer emits -- a nine-element tuple is nine items.
+                // Not `[* X]`, which would say any length of one type.
+                case TypeKind.Tuple:
+                {
+                    List<string> elements = new List<string>();
+
+                    foreach (ITypeSymbol element in TypeCollector.FlattenTupleElements((INamedTypeSymbol)type))
+                    {
+                        string? renderedElement = Render(element, byKey, ruleNames, options, Position.Value);
+
+                        if (renderedElement is null)
+                        {
+                            return null;
+                        }
+
+                        elements.Add(renderedElement);
+                    }
+
+                    rendered = "[" + string.Join(", ", elements) + "]";
+                    break;
+                }
+
                 default:
                     return null;
             }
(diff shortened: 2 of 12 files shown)


mcatanzariti (commented):
Yes, this supersedes #210 — I checked rather than took your word for it, and the flat encoding is ratified below. One thing has to happen before I close that PR, and it is not a change to this code.

## Verified

The decisive experiment for "supersedes": I took #210's three tests **verbatim** and ran them against this branch. All three pass. The trailing-break check does survive the move into `TupleReader.Read`, and the probes I ran against #210 — indefinite-length round trip through `ArrayLengthMode.IndefiniteLength`, a tuple in an indefinite-length map, a nested tuple, a `(int, int)?` member, a truncated `9F0102` — pass here too.

The rest holds up. Both TFMs green (`CDDL_REQUIRED=1`), 35 generator tests green, four TFMs build. Probing arities directly: 7 is byte-for-byte what it was (`8701020304050607`), and 8, 9, 10, 15, 16 and 22 are flat, with 22 — three levels of `Rest` — round-tripping. A wrong arity is refused in both length forms, and an item that fails past seven names `$[7]`, flattened, rather than a path through a `Rest` the document knows nothing about.

**The Native AOT claim I reproduced independently**, since it is the one thing no amount of reading settles: a probe published with `PublishAot`, run as a stripped ELF binary with no CoreCLR anywhere, over a context carrying arity 2, 9, 15 and a nullable tuple.

```
written : A4645061697282016374776F644E696E6589010203040506070809674669667465656E8F01…0F684F7074696F6E616C820765736576656E
ROUND TRIP OK
```

Byte-identical to the reflection path on CoreCLR. It works.

## Before #210 closes: bring its tests with it

**None of #210's three tests are here** — zero of three. What this PR adds covers arity past seven (`AWrongArityPastSevenIsRefused`, `AnIndefiniteLengthTuplePastSevenIsReadAndLeavesNothingBehind`, and the rest), which is the new ground, and that is the right instinct for a PR closing #198. But the fix for #199 also applies at arities two through seven, and after this lands nothing asserts it there: not the too-many-items case at each arity, not the tuple as a map member, not the tuple as an array element.

That coverage is the reason #199 was found and the reason it stays fixed. Please port the three tests as they stand — they pass unmodified, I ran them — and then I will close #210 rather than merge it.

While you are at it, the round trip through this library's own writer deserves to be one of them:

```csharp
var options = new CborOptions { ArrayLengthMode = LengthMode.IndefiniteLength };
string hex = Helper.Write(new List<(int, int)> { (1, 2), (3, 4) }, options);  // 9F9F0102FF9F0304FFFF
Cbor.Deserialize<List<(int, int)>>(hex.HexToBytes());
```

That was the sharpest statement of #199 — the library could not read back what it had just written — and it is not asserted in either PR.

## The wire format: flat, and that is now decided

You asked twice and I did not answer; you investigated and answered it yourself, and you got it right. Recording the decision here so it is not implicit in a merge: **a tuple is its elements in order, at any arity, flat.** A nine-element tuple is nine items. The seven-plus-`Rest` nesting is how C# spells a tuple, and a wire format has no business carrying another language's representation choice — a reader in Rust or Python sees an array of nine, which is the only thing it could reasonably see.

Two facts made this free, and both check out: arity seven and below is unchanged byte for byte, and nothing above seven ever worked, so there is no document anywhere that this PR reinterprets.

## One defect

`A tuple's Rest must itself be a tuple` is thrown from `Tuple8Converter`'s constructor, which the provider reaches through `Activator.CreateInstance`. So it arrives wrapped:

```
TargetInvocationException: Exception has been thrown by the target of an invocation.
  inner: CborException: A tuple's Rest must itself be a tuple, and System.DateTime is not
```

Reproduced with `ValueTuple<int, int, int, int, int, int, int, DateTime>`, which is legal C# and can be a member type. The message is exactly right; the envelope is the same one #198 complains about, so a caller following the `catch (CborException)` this library trains them into still misses it.

Not yours to fix here — `CborConverterProviderBase.CreateConverter` wraps every converter's constructor the same way, collections and nullables included, so unwrapping it with `ExceptionDispatchInfo` is a three-line change that belongs in its own PR. Worth opening; it makes the catchability half of #198 true rather than nearly true.

## Scope

The two commits are clean and each stands on its own, as you say. The second one is a feature — the generator learning a type kind, plus user-facing README — and by your own reading on #198 it is "a generator feature and its own issue". I am taking it here rather than asking you to split it, because it is the same design decision applied twice and splitting it would mean stating the recursion twice, but open the issue retroactively so the AOT support has something to be found by.

The generator side reads well, and the two details that would have been easy to get wrong are handled: detection by metadata name rather than `IsTupleType`, which is false for the one-element `ValueTuple` that every arity past seven bottoms out in, and the `Rest` registered before the converter that asks the registry for it. `FlattenTupleElements` giving CDDL the wire shape rather than the C# shape is the same call made consistently.

Two smaller notes, neither blocking. `Tuple8Converter` changes the type it converts, from `(T1, …, T8)` to `ValueTuple<T1, …, T7, TRest>`, and `Tuple1Converter` is new public surface — no real consequence, since the old class could not be constructed, but it is a public-API change and belongs in the release notes next to the format. And `case 1:` sits after `case 8:` in the provider; the comment explains why it exists, but reading order is arity order.

---

Issue #135, opened 2025-01-23
More meaningful error message


Odonno:
The current reader throws an error when it finds mismatched byte in the byte array. When this happens, we got an error along with a concrete StackTrace. The error generated could look like this:

```
CborException: [129] Expected major type TextString (3)
```

While it is helpful to detect which byte could be the source of the problem, it does not give much information about where in the POCO model the issue happens. I am curious to know if we could offer an error message that look like one of these:

```
CborException: [129] Expected major type TextString (3). Failed to deserialize at root level.
```

```
CborException: [129] Expected major type TextString (3). Failed to deserialize from "Array[7].Object.PropertyA".
```

This obsviously need the reader to traverse and collect the different levels, at least from object and array deserialization.

mcatanzariti:
Hello David,

Would you like to propose a pull request?

Michaël

Odonno:
I currently do not have the time to do this. Gonna get back to it when I can.

PR #217, opened 2026-08-14
Let a converter's own exception reach the caller, closes #216

### LINKED ISSUE #216: A CborException thrown from a converter's constructor reaches the caller as TargetInvocationException
Every converter this library builds through a provider is constructed with `Activator.CreateInstance`, so a `CborException` thrown from a converter's constructor reaches the caller wrapped in a `TargetInvocationException`. A `catch (CborException)` — which the rest of the API trains callers into — misses it.

Reproduced with `ValueTuple<int, int, int, int, int, int, int, System.DateTime>`, which is legal C# and can be a member type:

```
TargetInvocationException: Exception has been thrown by the target of an invocation.
  inner: CborException: A tuple's Rest must itself be a tuple, and System.DateTime is not
```

The message is right; the envelope is not. This is not specific to tuples — `CborConverterProviderBase.CreateConverter` wraps every converter's constructor the same way, collections and nullables included, so any constructor-time validation a converter does arrives the same way.

The fix is to unwrap and rethrow preserving the original stack:

```csharp
catch (TargetInvocationException exception) when (exception.InnerException is not null)
{
    ExceptionDispatchInfo.Capture(exception.InnerException).Throw();
    throw;   // unreachable; satisfies definite assignment
}
```

Worth noting it is a behaviour change for anyone currently catching `TargetInvocationException` and reading `.InnerException`, though catching that type deliberately seems unlikely given nothing documents it.

The test-side consequence is visible in this repository already: `IntDiscriminatorTests` carries an `AssertThrowsCborException` helper that walks `InnerException` for exactly this reason, and it could go away.

Raised out of the review of #214, where the maintainer suggested it belongs in its own PR rather than in that one. Happy to open one.

Rafael-SOWNet:
Closes #216.

Everything this library builds by reflection is built with `Activator.CreateInstance`, which wraps whatever the constructor threw in a `TargetInvocationException`. The exception was always the right one — only the envelope was wrong, and a caller's `catch (CborException)`, the one the rest of the API trains them into, matched none of them.

## What was wrapped, measured rather than reasoned about

Five places a constructor can run code that refuses. Each has a test in `Issues/Issue0216.cs`, and each failed before this change:

| Site | Reached by | Wrapped |
|---|---|--:|
| `CborConverterProviderBase.CreateConverter` | a converter built through any provider | once |
| `MemberMapping` → `IMemberConverter` | every member of every mapped type | once |
| `MemberMapping` → `[CborConverter]` | a converter the caller named on a member | **twice** |
| `ObjectMappingRegistry` → `ObjectMapping<T>` | any type whose mapping refuses itself | once |
| `DefaultObjectMappingConvention` → `[CborNamingConvention]` | a naming convention the caller named | **twice** |

The two double-wrapped ones are where the construction itself constructs something reflectively, so a single unwrap somewhere central would not have been enough. Each site unwraps its own layer instead.

The one an ordinary user reaches without writing a converter at all is the object mapping — two members under one CBOR name, which the mapping validation refuses:

```
TargetInvocationException: Exception has been thrown by the target of an invocation.
  inner: CborException: class/struct TwoMembersUnderOneName maps several fields/properties to the member name 'x'
```

`Util/TypeExtensions.GetDefaultValue` is routed through the same helper. A struct with a user-written parameterless constructor is legal C# and that constructor can throw.

## `ExceptionDispatchInfo`, not `throw exception.InnerException`

The latter resets the stack trace to the rethrow and loses the frame the refusal actually came from. `AnExceptionOfAnotherTypeIsNotConvertedOrSwallowed` pins both halves of that: an `InvalidOperationException` from a converter constructor arrives as itself rather than as a `TargetInvocationException` around it, **and** its own frame is still in `StackTrace`. Nothing is converted into a `CborException` on the way out.

## What this does not change

Nothing about which exception is thrown, or when. Every message is the one the constructor already produced. The only observable difference is the type a caller catches — so it is a behaviour change for anyone deliberately catching `TargetInvocationException` and reading `.InnerException`, which nothing documents and no test in this repository did.

## The workaround the test suite carried

Two copies of an `AssertThrowsCborException` helper walked `InnerException` for exactly this reason. Both are gone and their eleven call sites are plain `Assert.Throws<CborException>`.

Worth being precise about what that proves: **one of the eleven was genuinely wrapped** — `IntDiscriminatorTests.BothDiscriminatorAttributesThrows`, which refuses while the mapping is built, and which fails without this change. The nine in `TypedArrayTests` throw from `Read` rather than from a constructor and were never wrapped at all, so that helper's comment was wrong about its own call sites. Removing them is a simplification there, not evidence.

## Verification

**1902 library tests and 35 generator tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped; all four TFMs build. `ExceptionDispatchInfo` is available on netstandard2.0, so no target is left out.

**Six fail without the `src/Dahomey.Cbor/` change** — the five new tests plus `BothDiscriminatorAttributesThrows` — measured by stashing only the library paths and re-running.

---

### DIFF
--- a/src/Dahomey.Cbor.Tests/IntDiscriminatorTests.cs
+++ b/src/Dahomey.Cbor.Tests/IntDiscriminatorTests.cs
@@ -19,26 +19,6 @@ namespace Dahomey.Cbor.Tests
             return Cbor.Deserialize<T>(hexBuffer.HexToBytes(), options);
         }
 
-        /// <summary>
-        /// Converter construction happens through <see cref="Activator"/>, so a <see cref="CborException"/>
-        /// raised while building an object mapping surfaces wrapped in a <see cref="TargetInvocationException"/>.
-        /// </summary>
-        private static CborException AssertThrowsCborException(Action action)
-        {
-            Exception exception = Assert.ThrowsAny<Exception>(action);
-
-            for (Exception? current = exception; current != null; current = current.InnerException)
-            {
-                if (current is CborException cborException)
-                {
-                    return cborException;
-                }
-            }
-
-            throw new Xunit.Sdk.XunitException(
-                $"Expected a {nameof(CborException)} somewhere in the exception chain, got: {exception}");
-        }
-
         public interface IIntBaseInterface
         {
             int Id { get; set; }
@@ -369,7 +349,7 @@ namespace Dahomey.Cbor.Tests
         {
             CborOptions options = new CborOptions();
 
-            CborException exception = AssertThrowsCborException(
+            CborException exception = Assert.Throws<CborException>(
                 () => Helper.Write<IntBaseObject>(new BothDiscriminatorsObject(), options));
 
             Assert.Contains(nameof(CborDiscriminatorAttribute), exception.Message);
@@ -449,7 +429,7 @@ namespace Dahomey.Cbor.Tests
             // {"_t": "StringDisc", "Name": "foo", "Id": 1} read through the int-claimed base type
             const string hexBuffer = "A3625F746A537472696E6744697363644E616D6563666F6F62496401";
 
-            CborException exception = AssertThrowsCborException(
+            CborException exception = Assert.Throws<CborException>(
                 () => Deserialize<IntBaseObject>(hexBuffer, options));
 
             Assert.Contains("Invalid major type", exception.Message);
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0216.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0216.cs
@@ -1,0 +1,190 @@
+using Dahomey.Cbor.Attributes;
+using Dahomey.Cbor.Serialization;
+using Dahomey.Cbor.Serialization.Conventions;
+using Dahomey.Cbor.Serialization.Converters;
+using Dahomey.Cbor.Serialization.Converters.Providers;
+using System;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests.Issues
+{
+    /// <summary>
+    /// Everything this library builds by reflection is built with <c>Activator.CreateInstance</c>, which
+    /// wraps whatever the constructor threw in a <see cref="System.Reflection.TargetInvocationException"/>.
+    /// A caller's <c>catch (CborException)</c> — the one the rest of the API trains them into — misses
+    /// every one of these.
+    /// </summary>
+    /// <remarks>
+    /// The point is the envelope, not the message: each of these threw the right <c>CborException</c> all
+    /// along, and it arrived as something else. The cases below are the four places a constructor can run
+    /// code that fails — the converter a provider builds, a member's own <c>[CborConverter]</c>, a type's
+    /// object mapping, and a <c>[CborNamingConvention]</c> — which is every reflective construction site
+    /// where the thing being constructed can refuse.
+    /// </remarks>
+    public class Issue0216
+    {
+        /// <summary>
+        /// A converter built through a provider. <c>CborConverterProviderBase</c> is public and
+        /// <c>CreateConverter</c> is protected, so this is the supported extension point rather than a
+        /// contrivance; in-library converters reach it the same way.
+        /// </summary>
+        [Fact]
+        public void AConverterRefusingItsTypeThrowsCborException()
+        {
+            CborOptions options = new CborOptions();
+            options.Registry.ConverterRegistry.RegisterConverterProvider(new RefusingConverterProvider());
+
+            CborException exception = Assert.Throws<CborException>(
+                () => Helper.Write(new RefusedByItsConverter(), options));
+
+            Assert.Equal("this type is refused by its converter", exception.Message);
+        }
+
+        /// <summary>A member's own <c>[CborConverter]</c>, which is a type the caller named.</summary>
+        [Fact]
+        public void AMemberConverterRefusingItsTypeThrowsCborException()
+        {
+            CborException exception = Assert.Throws<CborException>(
+                () => Helper.Write(new HolderWithARefusingMemberConverter()));
+
+            Assert.Equal("this member is refused by its converter", exception.Message);
+        }
+
+        /// <summary>
+        /// An object mapping that refuses itself while being built. This is the one an ordinary user
+        /// reaches without writing a converter at all: two members under one CBOR name, which the mapping
+        /// validation refuses.
+        /// </summary>
+        [Fact]
+        public void AMappingRefusingItselfThrowsCborException()
+        {
+            CborException exception = Assert.Throws<CborException>(
+                () => Helper.Write(new TwoMembersUnderOneName()));
+
+            Assert.Contains("TwoMembersUnderOneName", exception.Message);
+        }
+
+        /// <summary>A <c>[CborNamingConvention]</c> whose constructor refuses.</summary>
+        [Fact]
+        public void ANamingConventionRefusingItselfThrowsCborException()
+        {
+            CborException exception = Assert.Throws<CborException>(
+                () => Helper.Write(new HolderWithARefusingNamingConvention()));
+
+            Assert.Equal("this naming convention is refused", exception.Message);
+        }
+
+        /// <summary>
+        /// The unwrap must not swallow anything: an exception that is not a <c>CborException</c> reaches
+        /// the caller as itself rather than as a <c>TargetInvocationException</c> around it, and keeps the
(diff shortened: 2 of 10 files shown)


mcatanzariti:
Reviewed at 5db3185, built and run locally on net8.0/net10.0 with `CDDL_REQUIRED=1`.

## Verified

| | |
|---|---|
| Build, four TFMs | 0 errors |
| Library tests, `net10.0` | 1902 / 1902 |
| Library tests, `net8.0` | 1902 / 1902 |
| Generator tests, `net10.0` | 35 / 35 |
| With `src/Dahomey.Cbor/` reverted to c79a39f | exactly 6 failures — the 5 new ones plus `IntDiscriminatorTests.BothDiscriminatorAttributesThrows` |
| Merge onto `master` (2 commits ahead: #208, #214) | clean, despite both touching `TypedArrayTests.cs` |

Every claim in the commit message checks out, including the admission that 9 of the 11 old
`AssertThrowsCborException` call sites were never wrapped and the comment saying so was wrong.

## What the change gets right

`ExceptionDispatchInfo` rather than `throw exception.InnerException`, and the choice is pinned by a
test rather than asserted in a comment: `AnExceptionOfAnotherTypeIsNotConvertedOrSwallowed` checks
that the `ThrowingConverter` frame is still in `StackTrace`. That is the detail this class of fix
usually loses.

One unwrap per site rather than a `while (InnerException != null)` walk is the right call too. It
composes correctly under nesting — the inner site rethrows bare, the outer constructor re-wraps once,
the outer site unwraps once — and it cannot strip an envelope a caller deliberately put on.

And a foreign exception arrives as itself. The library does not annex other people's exception types.

## `CreatorMapping` is not covered, and the comment says it is

`ReflectionActivator`'s remarks claim to cover "every reflective construction site where the thing
being constructed can refuse". `CreatorMapping.CreateInstance` builds the object through
`Delegate.DynamicInvoke` (`CreatorMapping.cs:103` and `:127`), and `DynamicInvoke` wraps exactly the
way `Activator` does.

Probed on this branch with a `[CborConstructor]` that throws a `CborException`, deserializing
`A16249640C` (`{"Id": 12}`):

```
OBSERVED: System.Reflection.TargetInvocationException / inner=Dahomey.Cbor.CborException
```

So #216 survives untouched on that path — and it is the one closest to an ordinary user of the six:
a constructor the user wrote, running at read time rather than while a mapping is built. Both
`MapCreator(ConstructorInfo)` and `MapCreator(MethodInfo)` reach it.

Either extend the fix to those two call sites, or invoke the delegate in a typed way instead of
through `DynamicInvoke` — which removes the wrapping and the late-bound dispatch cost at once. If
neither is in scope here, the remarks should be narrowed, because as written they tell the next
reader the subject is closed.

## Minor

- `CreateInstance(Type, params object?[] arguments)`: no caller passes `null` or a pre-built
  `object?[]` today, but under `params` both would silently change meaning (a `null` becomes "no
  arguments"). A non-`params` overload, or a line of comment, would hold that shut.
- `MemberMapping.cs:94` (a value type's default) is correct, but it is the only one of the five with
  no test of its own. It only bites on a struct with an explicit parameterless constructor, so this
  is a note rather than a request.
- The exception type callers observe changes, which is externally visible. No CHANGELOG in the repo,
  so nothing to update, but it is worth a release-note line.

## Verdict

Approvable. The fix is correct, honestly described, and rebases without friction. I would want either
the two `DynamicInvoke` sites handled or the exhaustiveness claim in the comment corrected before
merge.

Rafael-SOWNet:
You are right, and the comment was the worst part of it — it told the next reader the subject was closed. Fixed at `4693dcb`.

## `CreatorMapping` reproduced before agreeing

Both routes, and they are now tests rather than a claim:

```
ACreatorConstructorRefusingItsArgumentsThrowsCborException   [CborConstructor]
ACreatorFactoryRefusingItsArgumentsThrowsCborException       MapCreator(o => Factory.Create(o.Id))
```

Against the library at `c79a39f` both give `TargetInvocationException / inner=CborException`, exactly as you observed. You are also right that this is the closest of the six to an ordinary user: nothing in it is a converter, a provider or a convention — it is a constructor on their own type, refusing the values a document carried, at read time.

**The envelope cost a second thing there, which I had not seen until the test passed.** `CborException.Path` is prepended by a `catch (CborException)` on the way out, and a `TargetInvocationException` matched none of those frames — so a creator's refusal arrived with **no path at all**. It now reports `$`, and the test asserts it:

```csharp
Assert.StartsWith("this creator refuses 12", exception.Message);
```

So on this path #216 and #135 were the same defect wearing one envelope.

## Unwrap rather than typed invoke, and why

I took the first of your two options. Compiling the creator to a `Func<object?[], object>` is the better end state — it removes the wrapping and the late-bound dispatch together, and I would like to do it — but it changes *how creators are invoked*, and that does not belong inside a fix about which exception type a caller catches. It also has a wrinkle worth settling on its own: `CreatorMapping(IObjectMapping, Delegate)` takes a delegate the caller supplied, so the adapter has to be built by `Expression.Invoke` over a constant rather than from a `ConstructorInfo`, and that is a design question rather than a mechanical change.

Happy to open it as its own PR against a green master if you want it; say the word.

## The remarks now say what they cover

They no longer claim every route into user code, because that claim cannot be made true — a converter's `Read` and `Write` are called directly and were never wrapped, and neither is an `Expression`-compiled member accessor. They now name `Activator` and `DynamicInvoke` and stop there.

I also kept your framing about composition, since it is the argument for one unwrap per site rather than a walk, and it was not written down anywhere.

## The two minors

- **`params` is gone.** You are right that `null` and a pre-built `object?[]` both change meaning silently and neither reads differently at the call site. `CreateInstance(Type, object?[])` is now a plain overload and all three callers spell the array out.
- **`MemberMapping.cs:94`** — left as it is, and I agree it is a note rather than a request. It only bites on a struct with an explicit parameterless constructor that throws, and a test for it would pin a case the library has no other opinion about. It goes through the same helper, so it behaves the same way; I would rather not add a fixture that exists only to be exhaustive.
- **Release note** — agreed, and there is nowhere in the repo to put it. Worth saying here for the record: **the exception type callers observe changes on six paths.** Anyone catching `TargetInvocationException` and reading `.InnerException` around a `Cbor.Deserialize` or a context construction now catches nothing; `catch (CborException)` starts working where it did not.

## Verification

**1904 library tests and 35 generator tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped; four TFMs build.

**Eight fail against the library at `c79a39f`** — the seven new tests plus `IntDiscriminatorTests.BothDiscriminatorAttributesThrows` — measured by checking out that revision's `src/Dahomey.Cbor/` under the new tests, not by stashing, since the first commit is already in this branch's history.

---

mcatanzariti:
All good. Re-verified at `4693dcb`, built and run in Release with `CDDL_REQUIRED=1`.

| | |
|---|--:|
| Build, four TFMs incl. `netstandard2.0` | 0 errors |
| Library tests, `net10.0` / `net8.0` | 1904 / 1904 each, 0 skipped |
| Generator tests, `net10.0` | 35 / 35 |
| `Issue0216` in Release (the `StackTrace` assertion) | 7 / 7 |
| With `src/Dahomey.Cbor/` reverted to `c79a39f` | exactly 8 failures — the 7 new ones plus `IntDiscriminatorTests.BothDiscriminatorAttributesThrows` |
| Merge onto current `master` (`3b637ef`) | clean |

`CreatorMapping` is handled, and the `Path` finding that came with it is the better half of that second commit: #216 and #135 were the same defect wearing one envelope on that path, which neither of us had seen. The remarks now say what they cover and stop there.

Agreed on leaving the typed-invoke rewrite for its own PR — it changes how creators are invoked, which is not what this change is about. Happy to see it against a green master whenever you want to pick it up.

## One thing, and it is not a blocker

Unwrapping changes what the library's *own* catches match, not only the caller's. I audited all fifteen: fourteen rethrow after enriching `Path`, and one swallows — `TryReadItem` in `ReadNextItemAsync`, where a `CborException` means "wait for more bytes". A refusing `[CborConstructor]` now lands there, so on a completed pipe its message is replaced by `"Reader has completed with some buffer bytes but no item was read"`. Measured against `c79a39f`, where it arrived as a `TargetInvocationException` with the real message inside.

The conflation predates this PR — any `CborException` from a `Read` already vanished the same way on that path — so it is not yours to fix here. Filed as #219 with the repro and a three-line suggestion.

Merging this.


---

PR #218, opened 2026-08-14
Exclude a CDDL comment from the rule-name scan by its own syntax

Rafael-SOWNet:
The non-blocking nit from your #208 approval, as its own change since that PR has merged.

`RuleNames` treated any line containing `" = "` that did not start with a space as a rule. That is correct today only because no header comment happens to contain one — so it rests on the current header's wording rather than on CDDL's syntax. Excluding a comment by its leading `;` closes it:

```csharp
if (line.StartsWith(" ", StringComparison.Ordinal)
    || line.StartsWith(";", StringComparison.Ordinal))
{
    continue;
}
```

**No behaviour change today, and none is testable** — the condition cannot arise from the emitter as it stands, which is exactly why it would go unnoticed if a header line ever gained an `=`. What it prevents is `EveryCharacterOutsideAsciiIsEscapedInARuleName` failing on a comment it scanned as a rule name, which would point at the escaping rather than at the line that actually changed.

117 CDDL tests green on net10.0 with `CDDL_REQUIRED=1` and gem 0.12.14; all four TFMs build.

Take or leave — it is a test-helper robustness change and nothing depends on it.

---

### DIFF
--- a/src/Dahomey.Cbor.Tests/Cddl/CddlIdentifierSyntaxTests.cs
+++ b/src/Dahomey.Cbor.Tests/Cddl/CddlIdentifierSyntaxTests.cs
@@ -96,13 +96,25 @@ namespace Dahomey.Cbor.Tests.Cddl
         }
 
         /// <summary>The left-hand side of every rule in the schema.</summary>
+        /// <remarks>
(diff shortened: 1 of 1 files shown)


mcatanzariti:

PR #220, opened 2026-08-15
Report the failure that stopped a sequence read, closes #219

### LINKED ISSUE #219: A creator's refusal loses its message on the PipeReader sequence path
Split out of the review of #217, which is otherwise good to merge.

## What happens

`Cbor.ReadNextItemAsync<TItem>(PipeReader)` reads speculatively: its local `TryReadItem` catches `CborException` and returns `false`, which the loop reads as "not enough bytes yet, wait for more" (`Cbor.cs:625-635`). When the pipe is already complete, the loop then throws a generic message of its own.

Since #217 unwraps `TargetInvocationException` at every reflective construction site, a `[CborConstructor]` that refuses now raises a bare `CborException` — so it lands in that catch and is read as a truncated stream rather than as a refusal.

Measured both ways, with a creator throwing `CborException($"this creator refuses {id}")`, deserializing `A16249640C` (`{"Id": 12}`) through a completed pipe:

```
before #217 (c79a39f) : TargetInvocationException: Exception has been thrown by the target of an invocation.
                          inner=CborException: this creator refuses 12      <- wrong type, message present

after #217 (4693dcb)  : CborException: Reader has completed with some buffer bytes but no item was read
                                                                            <- right type, message gone
```

`Cbor.Deserialize<T>` on the same bytes reports it correctly — `this creator refuses 12. Failed to deserialize from "$".` — so this is specific to the sequence API.

On a pipe that is *not* complete, both versions loop waiting for more data. Only the completed-pipe case changed.

## Scope

This is the only site in the library that swallows a `CborException`; the other fourteen catches rethrow after enriching `CborException.Path`. So the blast radius is exactly `ReadNextItemAsync`.

Worth being precise about what #217 did and did not cause: conflating "incomplete data" with "a genuine refusal" predates it — any `CborException` thrown from a converter's `Read` already disappeared the same way on this path. #217 moved creator refusals into that category. It is a widening of an existing conflation rather than a new mechanism, which is why it belongs here rather than in that PR.

## Suggested fix

Keep the caught `CborException` in `TryReadItem` and rethrow it — `ExceptionDispatchInfo`, to preserve the frame — when `result.IsCompleted`, instead of throwing `"Reader has completed with some buffer bytes but no item was read"`. Passing it as `innerException` of that message would also do. Either way the "wait for more bytes" behaviour on an open pipe is unchanged, and a refusal keeps its message once the pipe closes.

A test on `ReadNextItemAsync` with a refusing creator over a completed pipe would pin it; there is none today.

mcatanzariti:
Closes #219.

## What was wrong

`Cbor.ReadNextItemAsync(PipeReader)` reads speculatively: an item that stops halfway through the bytes available so far cannot be told apart from one whose remaining bytes have not arrived yet, so its local `TryReadItem` catches `CborException` and answers "not yet". On an open pipe that is the only honest answer.

On a completed pipe there is nothing left to wait for — and the method was discarding what it had caught and throwing a message of its own:

```
CborException: Reader has completed with some buffer bytes but no item was read
```

So a converter that refuses what it read was reported as a truncated stream. `Cbor.Deserialize` on the same bytes has always reported it correctly; only the sequence API lost it.

Worth being precise about what #217 did and did not cause. Conflating "incomplete data" with "a genuine refusal" predates it: a `CborException` thrown from a converter's `Read` already disappeared here. #217 unwrapped `TargetInvocationException` at the reflective construction sites, which moved `[CborConstructor]` refusals out of an envelope this catch did not match and into one it did — a widening of an existing conflation rather than a new mechanism.

## The fix for #219

The caught exception is rethrown through `ExceptionDispatchInfo` when `result.IsCompleted`, rather than replaced. A refusal keeps its message, its `CborException.Path` and the frame it was thrown from; a genuinely truncated document now reports the reader's own `[N] Unexpected end of buffer` instead of a message invented one frame up. The open-pipe path is untouched — `TryReadItem` still returns `false` and the loop still waits.

The speculative read also now goes through `RootReader`, so a failure on the root value itself is placed at `$` exactly as `Cbor.Deserialize` places it. Without it this branch handed back `Path == null`, which contradicts what `CborException.Path` documents.

## What review turned up on the way

Restoring the exception exposed the state the method left the `PipeReader` in, which had not mattered while every failure here was terminal. Four rounds of review, each finding reproduced before being acted on:

* **The read was never ended on the failing routes.** A `PipeReader` left between a `ReadAsync` and its `AdvanceTo` answers every later call with `InvalidOperationException: Reading is already in progress` — so the exception this PR restores was readable exactly once and masked from then on. This reached the caller by three separate doors: the `CborException` route, a converter throwing something else (supported since #216), and the end of the sequence, which is the method's most common exit. A `finally` guarded by a flag now covers every route; the success path advances past what it consumed, everything else advances past nothing.
* **`examined` ran to `buffer.End` on the success path** while only one item was consumed, telling the pipe this read had already seen the rest. On an open `Pipe` carrying two items in one flush, the second read never returned. `examined` now stops where `consumed` stops.
* **The `IsCanceled` route** skipped the read-ending path entirely and misreported: a cancelled result is not always a cancelled token — `PipeReader.CancelPendingRead` cancels the read alone — and handing that token to `Task.FromCanceled` raised `ArgumentOutOfRangeException` about a parameter the caller never passed. It now throws `OperationCanceledException`, and the reader survives it.
* **`ExceptionDispatchInfo.Capture` ran on every speculative attempt**, including the thousands discarded while a fragmented item arrives, though only the completed branch has any use for one. The capture moved to that branch.

The method's remarks now state what a caller has to know rather than leaving it in a test: a failure consumes nothing, so calling again re-reads the same bytes and raises the same failure. There is no skip-and-continue here — a throw is the end of the sequence. Skipping would mean deciding how many bytes a failed read is deemed to have eaten, which is a question for an API that offers to continue.

Every exit of the method now passes through exactly one `AdvanceTo`: cancellation, empty buffer, success, refusal on a completed pipe, refusal on an open one, a non-`CborException` either way, and a mapping the registry refuses to build.

## Tests

`Issues/Issue0219.cs`, ten cases. The four for #219 itself fail against 2587151; the rest fail against the commit each was written for.

* a `[CborConstructor]` refusal over a completed pipe — message, path (`$`) and origin frame all survive
* a converter's `Read` refusing — the older half of the same conflation, which never involved a `TargetInvocationException`
* a failure on the root value itself, asserted equal to `Cbor.Deserialize`'s own answer rather than to a literal, so the two cannot drift apart
* a genuinely truncated item — still a `CborException`, now in the reader's own words
* an open pipe — the read holds rather than reporting anything, and the refusal surfaces when the writer completes, not before
* the reader still usable after a `CborException`, after an exception of another type, and at the end of the sequence
* two items flushed together on an open pipe, both readable
* a read cancelled through `CancelPendingRead`

1953 library tests green on net10.0 and net8.0, 35 generator tests, four TFMs building.

### DIFF
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0219.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0219.cs
@@ -5,6 +5,7 @@ using Dahomey.Cbor.Tests.Extensions;
 using System;
 using System.Buffers;
 using System.IO.Pipelines;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
@@ -46,8 +47,10 @@ namespace Dahomey.Cbor.Tests.Issues
             Assert.StartsWith("this creator refuses 12", exception.Message);
 
             // Rethrown rather than wrapped, so the path collected on the way out survives too - the
-            // same "$" Cbor.Deserialize reports for these bytes.
+            // same "$" Cbor.Deserialize reports for these bytes - and so does the frame it was thrown
+            // from, which is what rethrowing through ExceptionDispatchInfo buys over `throw failure`.
             Assert.Equal("$", exception.Path);
+            Assert.Contains(nameof(RefusedByItsCreator), exception.StackTrace);
         }
 
         /// <summary>
@@ -213,6 +216,42 @@ namespace Dahomey.Cbor.Tests.Issues
             CborException exception = await Assert.ThrowsAsync<CborException>(() => reading);
 #pragma warning restore VSTHRD003
             Assert.StartsWith("this creator refuses 12", exception.Message);
+        }
+
+        /// <summary>
+        /// A read reports as examined only what it consumed. Reporting the rest of the buffer would
+        /// tell the pipe this read had already seen it, and the next read would wait for bytes beyond
+        /// it - so items already flushed behind this one would hang until the writer wrote more.
+        /// </summary>
+        [Fact]
+        public async Task ItemsAlreadyFlushedBehindThisOneAreReadableAsync()
+        {
+            Pipe pipe = new Pipe();
+            await pipe.Writer.WriteAsync("0C0D".HexToBytes());   // 12 then 13, in one flush, writer left open
+
+            using CancellationTokenSource timeout = new CancellationTokenSource(TimeSpan.FromSeconds(5));
+
+            Assert.Equal(12, await Cbor.ReadNextItemAsync<int?>(pipe.Reader, cancellationToken: timeout.Token));
+            Assert.Equal(13, await Cbor.ReadNextItemAsync<int?>(pipe.Reader, cancellationToken: timeout.Token));
+        }
+
+        /// <summary>
+        /// A cancelled read is not always a cancelled token: <see cref="PipeReader.CancelPendingRead"/>
+        /// cancels the read alone, and handing that token to <c>Task.FromCanceled</c> raised an
+        /// <see cref="ArgumentOutOfRangeException"/> about a parameter the caller never passed.
+        /// </summary>
+        [Fact]
+        public async Task ACancelledReadReportsAsCancelledAsync()
+        {
+            Pipe pipe = new Pipe();
+            await pipe.Writer.WriteAsync(OneWholeItem.HexToBytes());
+            pipe.Reader.CancelPendingRead();
+
+            await Assert.ThrowsAnyAsync<OperationCanceledException>(
+                async () => await Cbor.ReadNextItemAsync<Plain>(pipe.Reader));
+
+            // And the read it cancelled is over, so the reader still works.
+            Assert.Equal(12, (await Cbor.ReadNextItemAsync<Plain>(pipe.Reader))!.Id);
         }
 
         public class Plain
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0219.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0219.cs
@@ -96,6 +96,41 @@ namespace Dahomey.Cbor.Tests.Issues
         }
 
         /// <summary>
+        /// The read that failed is over, and the pipe knows it: a <see cref="PipeReader"/> left between
+        /// a <c>ReadAsync</c> and its <c>AdvanceTo</c> answers every later call with "Reading is already
+        /// in progress" instead of with the failure, so the exception this fix restores would be
+        /// readable exactly once and masked from then on.
+        /// </summary>
+        /// <remarks>
+        /// Nothing is consumed, so asking again asks the same question and gets the same answer rather
+        /// than moving past the item that refused. Skipping it would mean deciding how many bytes a
+        /// failed read is deemed to have eaten, which is a question for an API that offers to continue,
+        /// not for the one that reports what stopped.
+        /// <para>
+        /// A real <see cref="Pipe"/> rather than <see cref="PipeReader.Create(ReadOnlySequence{byte})"/>
+        /// because only the former tracks the read state this is about; the sequence-backed reader
+        /// re-answers either way, which is why the tests above never noticed.
+        /// </para>
+        /// </remarks>
+        [Fact]
+        public async Task TheReaderIsLeftUsableAfterTheFailureAsync()
+        {
+            Pipe pipe = new Pipe();
+            await pipe.Writer.WriteAsync(OneWholeItem.HexToBytes());
+            await pipe.Writer.CompleteAsync();
+
+            CborException first = await Assert.ThrowsAsync<CborException>(
+                async () => await Cbor.ReadNextItemAsync<RefusedByItsCreator>(pipe.Reader));
+
+            // Not InvalidOperationException, and not a hang: the same refusal, reported again.
+            CborException second = await Assert.ThrowsAsync<CborException>(
+                async () => await Cbor.ReadNextItemAsync<RefusedByItsCreator>(pipe.Reader));
+
+            Assert.StartsWith("this creator refuses 12", first.Message);
+            Assert.Equal(first.Message, second.Message);
+        }
+
+        /// <summary>
         /// A stream that really is truncated still fails, and now says so in the reader's own words
         /// rather than in a message invented one frame up. The distinction the fix restores is between
         /// two failures that were both reported as this one.
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0219.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0219.cs
@@ -1,0 +1,146 @@
+using Dahomey.Cbor.Attributes;
+using Dahomey.Cbor.Serialization;
+using Dahomey.Cbor.Serialization.Converters;
+using Dahomey.Cbor.Tests.Extensions;
+using System;
+using System.Buffers;
+using System.IO.Pipelines;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests.Issues
+{
+    /// <summary>
+    /// <c>ReadNextItemAsync</c> reads speculatively: an item that stops halfway through the bytes
+    /// available so far cannot be told apart from one whose remaining bytes have not arrived yet, so a
+    /// <see cref="CborException"/> means "retry on more data" while the pipe is open. Once the pipe is
+    /// complete there is no more data to wait for, and the failure that stopped the last attempt was
+    /// being replaced by a message of the method's own - so a converter's refusal was reported as a
+    /// truncated stream.
+    /// </summary>
+    /// <remarks>
+    /// The conflation predates #217: any <c>CborException</c> raised from a converter's <c>Read</c>
+    /// disappeared the same way. #217 widened it to <c>[CborConstructor]</c> refusals, which until then
+    /// arrived wrapped in a <c>TargetInvocationException</c> and so escaped the catch by accident. Both
+    /// routes are pinned below, and this is the only site in the library that swallows a
+    /// <c>CborException</c> - the other catches rethrow after enriching <see cref="CborException.Path"/>.
+    /// </remarks>
+    public class Issue0219
+    {
+        /// <summary>{"Id": 12}, a whole item - nothing about these bytes is truncated.</summary>
+        private const string OneWholeItem = "A16249640C";
+
+        /// <summary>
+        /// The case the issue is about: a creator that refuses the values the document carried, read
+        /// through a pipe that is already complete. <c>Cbor.Deserialize</c> on the same bytes always
+        /// reported this correctly; only the sequence API lost it.
+        /// </summary>
+        [Fact]
+        public async Task ACreatorRefusalReachesTheCallerOnACompletedPipeAsync()
+        {
+            PipeReader pipeReader = PipeReader.Create(new ReadOnlySequence<byte>(OneWholeItem.HexToBytes()));
+
+            CborException exception = await Assert.ThrowsAsync<CborException>(
+                async () => await Cbor.ReadNextItemAsync<RefusedByItsCreator>(pipeReader));
+
+            Assert.StartsWith("this creator refuses 12", exception.Message);
+
+            // Rethrown rather than wrapped, so the path the converters enriched on the way out survives
+            // too - the same "$" Cbor.Deserialize reports for these bytes.
+            Assert.Equal("$", exception.Path);
+        }
+
+        /// <summary>
+        /// The older half of the same conflation: a converter's <c>Read</c> refusing outright. This one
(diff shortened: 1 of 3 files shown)


mcatanzariti:
## Code review of this PR, and what came of it

Three findings. One was real and is fixed in d32693d; two are recorded and not taken, with the reasoning below.

### Fixed — the failure reached the caller with no path

`TryReadItem` called `converter.Read` directly rather than `RootReader.Read`, so the exception this PR now hands to the caller arrived with `Path == null` where `Cbor.Deserialize` on the same bytes reports `"$"`. Measured on `A16249640C` (a map) asked for as an `int`:

```
Deserialize   path=[$]      [1] Invalid major type Map. Failed to deserialize from "$".
ReadNextItem  path=[<null>] [1] Invalid major type Map
```

That contradicts what `CborException.Path` documents — a null value means the exception did not come from a read at all. It also made the justification written into `RootReader`'s own remarks false: it turns on the speculative read *discarding what it catches*, which as of the first commit here it no longer always does. Those remarks now say what is true of this path.

The two tests already in the PR passed either way, which is why this got through: both decode objects, and `ObjectConverter` marks the path itself on the way out. A primitive at the root passes through no such converter, so the added case reads that — and asserts equality with `Cbor.Deserialize`'s own answer in both path and message rather than against a literal, so the two cannot drift apart. The converter-refusal test no longer freezes a pathless message.

Marking a path on a failure that is about to be discarded costs nothing: it sets a bool on an exception the speculative loop was already constructing.

### Not taken — `Lookup` sits outside `TryReadItem`'s `try`

True as stated: a mapping that refuses while being built throws from `ConverterRegistry.Lookup`, one line above the `try`, and so is never captured in `failure`.

Measured on a type mapping two properties to one name:

* **completed pipe** — the case this PR is about — it reaches the caller as itself already: `CborException: class/struct TwoUnderOneName maps several fields/properties to the member name 'x'`. Nothing here changes it, and nothing needs to.
* **open pipe** — it hangs. But that is what *every* refusal does on an open pipe, and it is what the open-pipe test in this PR pins as intended. No amount of further bytes changes a refusal's mind, yet the loop cannot tell one from a truncation — that indistinguishability is the whole premise of reading speculatively.

Worth its own issue about lifting the lookup out of the loop, where the behaviour change can be argued on its own terms: hoisting it would also make `ReadNextItemAsync` on an empty completed pipe throw instead of returning `default`. Not something to smuggle in under a fix about which exception a caller sees.

The review reported this as reproducing with an interface lacking a `CreatorMapping`. That particular refusal comes from `ObjectConverter.Read` (`ObjectConverter.cs:324`), inside the `try` — so it demonstrates the ordinary open-pipe wait rather than the `Lookup` route. The finding is right; the case it was measured on isn't an instance of it.

### Not taken — `ExceptionDispatchInfo.Throw()` and reachability

`Throw()` is not annotated `[DoesNotReturn]` on every target, so the compiler no longer proves the rest of the `else` unreachable. Noted rather than worked around: the code after it is the open-pipe path, which is correct to fall into, and only unreachable by accident of the branch above throwing.

### Checked and dismissed

No `AdvanceTo` on the throwing path (identical to before), the cost of one `ExceptionDispatchInfo` per speculative attempt (negligible — no allocation on the success path), and the timing in the open-pipe test (deterministic in the direction that matters; it costs 250 ms).

1948 library tests green on net10.0 and net8.0, 35 generator tests, four TFMs building.

---

mcatanzariti:
## Second review pass — one finding, taken

The throwing branch returned without calling `AdvanceTo`, leaving the `PipeReader` between a `ReadAsync` and its `AdvanceTo`. Reproduced before agreeing, on a real `Pipe` holding two items:

```
#1  CborException: refuses 12. Failed to deserialize from "$".
#2  InvalidOperationException: Reading is already in progress.
```

So the exception the earlier commits restored was readable exactly once and masked from then on. Fixed in 9a3df4c: `AdvanceTo` now runs before the throw on both branches — which is also where it always ran on the waiting one. It consumes nothing, so the item that refused is left exactly where it is.

After the fix both reader kinds agree:

```
real pipe      #1 CborException: refuses 12…   #2 CborException: refuses 12…
PipeReader.Create  #1 CborException: refuses 12…   #2 CborException: refuses 12…
```

**Why this belongs here even though the throw site predates the PR.** The old message was terminal by nature — the stream is broken, there is nothing to resume. A creator's refusal is per-item, and a consumer that logs it and calls again is the obvious thing to write. The PR is what turns this branch from the former into the latter, so the reader state it leaves behind starts to matter.

**Reported again, not skipped.** Nothing was consumed, so asking again asks the same question and gets the same answer. Moving past the refused item would mean deciding how many bytes a failed read is deemed to have eaten — well defined for a creator that refused a whole item, not for a truncation — and that is a question for an API that offers to continue, not for one that reports what stopped. Worth raising separately if a skip-and-continue sequence read is wanted.

The new test uses a real `Pipe` rather than `PipeReader.Create`: only the former tracks the read state this is about, which is why none of the existing tests noticed.

1949 library tests green on net10.0 and net8.0, four TFMs building.

---

mcatanzariti:
## Third pass — both findings taken, 6e20bb4

My previous commit put `AdvanceTo` on the failing branch, which covers only the exceptions `TryReadItem` catches. Two routes still returned mid-read, both reproduced:

* **a converter throwing something other than a `CborException`** — supported since #216 — passes through `TryReadItem` untouched and past the `when (!result.IsCompleted)` filter, so nothing advanced: `#1 InvalidOperationException: not a CborException`, `#2 InvalidOperationException: Reading is already in progress`. The same masking, by the other door.
* **the end of the sequence** — the most common exit of the method. An empty buffer returned `default` without ending the read, so a consumer polling a drained pipe got one `null` and then `Reading is already in progress`.

A `finally` guarded by a flag now covers every route, which is what the invariant needed to be stated as rather than repeated per branch: the success path advances past what it consumed, everything else advances past nothing, and either way the read is over before the method is. The `catch` block's own `AdvanceTo` goes away, since the `finally` does it.

Three tests on this now — the `CborException` route, the other-exception route, and the drained pipe. The two new ones fail against the library at 9a3df4c.

1951 library tests green on net10.0 and net8.0, four TFMs building.

---

mcatanzariti:
## Fourth pass — all four taken, d0d8ccf

* **`examined: buffer.End` on the success path.** A read reported the whole buffer as examined while consuming only the item it read, so the next read waited for bytes beyond it. On an open `Pipe` carrying `12` and `13` in one flush, the second read returned only when a 2 s token cancelled it. Examined now stops where consumed stops. This is the sharpest of the four — a hang on the ordinary streaming case.
* **The `IsCanceled` route skipped the `try`/`finally`.** It neither ended the read nor reported usefully: a cancelled result is not always a cancelled token — `CancelPendingRead` cancels the read alone — and handing that token to `Task.FromCanceled` raised `ArgumentOutOfRangeException` about a parameter the caller never passed. Now advances and throws `OperationCanceledException`, and the reader survives it.
* **`Capture` on every speculative attempt.** Moved to the completed branch; `TryReadItem` hands back the `CborException` itself. The frame survives either way since nothing rethrows it in between — now pinned by asserting the creator's own frame is still in the trace.
* **The replay behaviour was only in a test's remarks.** It is in the method's own docs now: a failure consumes nothing, so calling again re-reads the same bytes and raises the same failure; there is no skip-and-continue, and a throw is the end of the sequence.

1953 library tests green on net10.0 and net8.0, 35 generator tests, four TFMs building.

---

Rafael-SOWNet:
Reviewed at `d0d8ccf` by building and running rather than reading, since this fixes a consequence of my #217.

## Verified

| | |
|---|---|
| Build, four TFMs | 0 errors |
| Library tests, `net10.0`, `CDDL_REQUIRED=1` | 1953 / 1953, 0 skipped |
| `Issues/Issue0219.cs` against `origin/master`'s `Cbor.cs` | **10 of 10 fail** |
| Position relative to `master` | 2 commits behind (#211, #221), MERGEABLE |

Every test binds, and one of them binds in a way worth calling out: `ItemsAlreadyFlushedBehindThisOneAreReadableAsync` fails as a **five-second hang**, not an assertion. So the `examined: buffer.End` bug is a genuine stall — a consumer with two items already flushed behind the first waits for bytes that will never come — rather than a latency nit. That is a more serious defect than the one the issue was about, and it was found on the way past.

**I checked your "only site" claim independently.** Thirteen `catch (CborException)` sites in `src/Dahomey.Cbor/`: two in `RootReader`, two in `AbstractCollectionConverter`, four in `ObjectConverter`, one each in `ArrayConverter`, `TypedArrayConverter`, `TupleConverter`, `AbstractDictionaryConverter` — every one of them rethrows after `PrependPath…`/`MarkPathKnown`. `Cbor.cs:631` was the only one that swallowed. The claim holds.

The `finally` invariant is the part I would have got wrong: a `PipeReader` left between `ReadAsync` and `AdvanceTo` answers every later call with "Reading is already in progress", so the failure reported once would be masked from then on. Ending the read on the throwing route too, and on cancellation, is what makes `TheReaderIsLeftUsableAfter…` possible at all.

## The description no longer matches the branch

Not a defect, but it is what a reviewer reads as the spec, and it is two commits stale: it says **four** tests where there are ten, and **1951** where the branch measures 1953. More substantially, it describes only the #219 fix — the `examined: end` correction and the cancellation handling are separately significant changes with their own tests, and someone reading the merge commit later will not find them mentioned.

## An addition, offered rather than pushed

The reason #219 existed for as long as it did is that the contract held at every entry point except one, and the tests were arranged **by mechanism** rather than by entry point — so nothing compared the doors against each other. `Issue0219.cs` fixes the case; it does not close the shape.

This does, and it is the test that would have caught #219 the moment #217 landed:

```csharp
[Theory]
[MemberData(nameof(EntryPoints))]
public void AConverterRefusalKeepsItsMessage(string name, Func<ReadOnlyMemory<byte>, CborOptions, object> read)
{
    CborOptions options = new CborOptions();
    options.Registry.ConverterRegistry.RegisterConverter(
        typeof(RefusedByItsConverter), new RefusingConverter());

    CborException exception = Assert.Throws<CborException>(() => read(WholeItem.HexToBytes(), options));

    Assert.StartsWith(ConverterMessage, exception.Message);

    // The messages an entry point would invent if it discarded the refusal.
    Assert.DoesNotContain("no item was read", exception.Message);
    Assert.DoesNotContain("Unexpected end of buffer", exception.Message);
}
```

over `Deserialize(ReadOnlySpan)`, `Deserialize(ReadOnlySequence)`, `DeserializeMultiple` for both, `DeserializeAsync(Stream)`, `DeserializeAsync(PipeReader)` and `ReadNextItemAsync(PipeReader)` — twelve cases with the `[CborConstructor]` half.

Measured both ways: **12/12 pass on this branch; on `master` exactly 2 fail**, and they are the two `ReadNextItemAsync` rows, both reporting `Reader has completed with some buffer bytes but no item was read`. So it isolates your fix precisely, and from then on a new entry point without a row is a visible omission rather than a silent one.

Happy to open it as a follow-up PR once this merges — it is a test-only file and it cannot pass before this lands. Say if you would rather have it in here, or not at all.

## Verdict

Approvable as it stands, and worth merging ahead of the description tidy-up rather than behind it. Thank you for splitting it out of #217 rather than holding that PR — and for being precise on the scope: you are right that #217 widened an existing conflation rather than introducing one, and it is a fair reading that my change is what made it reachable from a `[CborConstructor]`.

---

mcatanzariti:
Yes — open it as a follow-up. Test-only, already written and already measured, and it catches the class rather than the instance. There is nothing to weigh.

One request, and it is the reason I want it rather than a nicety: let the table be **where a new entry point is declared**, not only where the existing ones are checked. The value here is not that it re-finds #219 — that is fixed — it is that the twelfth row is load-bearing for the thirteenth. If adding a door means adding a row, the omission becomes visible the day it happens. If the table is only ever read, it will age exactly like the tests it is meant to backstop.

**On your recount, you are right and I was wrong.** I wrote "the other fourteen catches" in the commit message and the description, taking the number from the issue rather than counting it myself. It is twelve others, thirteen including the one that swallowed — your enumeration matches mine now that I have actually run it. The claim it supports is unaffected, but the number was not mine to repeat.

**On the hang:** your five-second observation is the sharpest statement of that defect anywhere in this thread. A failing assertion says a value was wrong; a test that hangs says a consumer with data already in the pipe waits for bytes that will never come. Worth keeping in mind that it surfaced only because restoring the exception made the surrounding contract observable — the streaming defect was older and quieter than the one the issue was about.

The description was rebuilt before the merge and now covers the whole branch — ten tests, the `examined` correction and the cancellation handling each described on their own terms, so the merge commit is readable without this thread.

Thank you for reviewing it by running it rather than by reading it. Three of the four rounds here landed on code the previous round had just touched; yours is the one that checked the claims instead.

---

Issue #225, opened 2026-08-15
DiscriminatorMapping.EnsureInitialize never sets _isInitialized, so its double-checked lock never latches


mcatanzariti:
Noticed while fixing #224. No user-visible symptom — filing it because the code reads as though it latches and does not, so the next person to touch it will assume a guarantee that is not there.

`DiscriminatorMapping<T>` declares `_isInitialized`, tests it twice in the classic double-checked pattern, and never assigns it:

https://github.com/dahomey-technologies/Dahomey.Cbor/blob/f429ba5/src/Dahomey.Cbor/Serialization/Converters/Mappings/DiscriminatorMapping.cs#L48-L64

Both inner and outer checks therefore always pass, and every `MemberName` read takes `lock (this)`, hits the registry, and allocates a fresh string through `Encoding.UTF8.GetString`.

It is the only one of the four mappings in that folder with the flag left dead — `ObjectMapping` (`:364`), `CreatorMapping` (`:269`) and `MemberMapping` (`:159`) all set it inside the lock. That asymmetry is what makes it look like an omission rather than a choice.

## Why it has no symptom today

The two consumers of `MemberName` are both in `ObjectMapping`'s duplicate-name validation:

https://github.com/dahomey-technologies/Dahomey.Cbor/blob/f429ba5/src/Dahomey.Cbor/Serialization/Converters/Mappings/ObjectMapping.cs#L374
https://github.com/dahomey-technologies/Dahomey.Cbor/blob/f429ba5/src/Dahomey.Cbor/Serialization/Converters/Mappings/ObjectMapping.cs#L394

That runs once per type, at mapping-initialization time, not per read or write. So the repeated lock and allocation cost a handful of operations per type over the lifetime of the options. This is a tidiness issue, not a performance one, and I would not have raised it on its own.

## Why it is not a one-line fix

Adding `_isInitialized = true;` inside the lock reintroduces exactly the bug #224 just removed. The convention resolves to `null` for a type whose subtypes are not yet registered, and the current code gets away with that only *because* it never latches: it re-resolves on each call, so a later registration reaches it. Latching would freeze `_memberName` at `null`.

So the fix has to latch on success rather than on having tried:

```csharp
if (discriminatorConvention != null)
{
    _memberName = Encoding.UTF8.GetString(discriminatorConvention.MemberName);
    _isInitialized = true;
}
```

[...]

Issue #223, opened 2026-08-15
Unregistered subtype reports a missing CreatorMapping rather than the registration the README prescribes


Rafael-SOWNet:
Reading a member declared as an abstract class or an interface, without having registered the concrete subtype, reports:

```
CborException: A CreatorMapping should be defined for interfaces or abstract classes (Shape)
```

The remedy it names is not the one the README prescribes. `README.md` under *Polymorphism* documents the fix as:

```csharp
options.Registry.DiscriminatorConventionRegistry.RegisterType<Circle>();
options.Registry.DiscriminatorConventionRegistry.RegisterType<Square>();
```

and that is what makes it work — confirmed on `de6acd3`:

```csharp
public abstract class Shape { }
[CborDiscriminator("circle")] public class Circle : Shape { public double Radius { get; set; } }
public class Holder { public Shape Shape { get; set; } }

string hex = Helper.Write(new Holder { Shape = new Circle { Radius = 1.5 } });
// A1 655368617065 A2 625F74 66636972636C65 66526164697573 F93E00   -- "_t": "circle" is present

Cbor.Deserialize<Holder>(bytes, new CborOptions());                    // the message above
Cbor.Deserialize<Holder>(bytes, optionsWithRegisterType);              // -> Circle
```

So the behaviour is correct and documented; only the message points elsewhere. A creator mapping *is* the right answer when the payload carries no discriminator at all, which is presumably why it reads that way — but when a discriminator was written and the subtype simply is not registered, it sends the reader to build a factory they do not need.

Two things would close it, and the second is the useful one:

* say both remedies, or
* distinguish the two cases: if the document carries a discriminator the type could not resolve, name `DiscriminatorConventionRegistry.RegisterType<T>()` and the discriminator value that was not recognised. That is actionable — it tells the caller exactly which subtype is missing.

Minor, and I would not have raised it except that the write side needs no registration at all, so the asymmetry is easy to walk into: the document looks complete, carries its `"_t"`, and the error talks about constructors.

Happy to open a PR if you want the second form.

mcatanzariti:
Reproduced on `f429ba5`, and your reading of it is right. Please do open the PR — option 2 is the one worth having.

Confirming the mechanism, since it decides where the fix goes. `ObjectConverter<Shape>` resolves its convention once, in the constructor:

[...]

PR #226, opened 2026-08-15
Let a subtype registered after the first read take effect, closes #224

### LINKED ISSUE #224: DiscriminatorConventionRegistry caches the negative result, so RegisterType after first use is a silent no-op
Split out of #223, where it turned up while checking that the remedy the error message names actually takes.

`RegisterType<TSubtype>()` has no effect if anything already asked the registry about the base type on the same `CborOptions`. It does not throw and does not warn — the next read fails exactly as it did before registration.

```csharp
public abstract class Shape { }
[CborDiscriminator("circle")] public class Circle : Shape { public double Radius { get; set; } }

CborOptions options = new CborOptions();
DiscriminatorConventionRegistry registry = options.Registry.DiscriminatorConventionRegistry;

registry.GetConvention(typeof(Shape));      // null -- nothing registered yet
registry.RegisterType<Circle>();
registry.GetConvention(typeof(Shape));      // still null

CborOptions fresh = new CborOptions();
fresh.Registry.DiscriminatorConventionRegistry.RegisterType<Circle>();
fresh.Registry.DiscriminatorConventionRegistry.GetConvention(typeof(Shape));
                                            // DefaultDiscriminatorConvention<string>
```

Confirmed on `f429ba5`.

## Why

`GetConvention` memoises through `GetOrAdd`, so a lookup that resolves nothing stores `null` for the base type:

https://github.com/dahomey-technologies/Dahomey.Cbor/blob/f429ba5/src/Dahomey.Cbor/Serialization/Conventions/DiscriminatorConventionRegistry.cs#L49-L52

`RegisterType(typeof(Circle))` then resolves the convention for `Circle` and propagates it up the hierarchy with `TryAdd`, which cannot displace the `null` already sitting under `Shape`:

https://github.com/dahomey-technologies/Dahomey.Cbor/blob/f429ba5/src/Dahomey.Cbor/Serialization/Conventions/DiscriminatorConventionRegistry.cs#L62-L82

So the entry is not merely stale, it is permanent for the lifetime of those options.

## Why it is worth more than a doc note

A first failed `Deserialize` is enough to poison the entry, and the failure it produces is the one #223 is about — a message that sends the caller off to register the subtype. Following that advice on the options already in hand reproduces the identical error, which reads as the advice being wrong rather than as an ordering constraint. Whatever wording #223 settles on inherits this: the remedy it names has to work when the caller applies it.

Ordinary use is unaffected — configure options once, then deserialize — so this is not urgent. It bites interactive diagnosis, and any code that registers subtypes lazily as they become known.

## Shape of a fix

Overwriting the base-type entry on registration is a small change:

```csharp
_conventionsByType.AddOrUpdate(currentType, convention, (_, existing) => existing ?? convention);
```

That alone does not finish the job. `ObjectConverter<T>` captures the convention in its constructor and the converter is itself cached in the converter registry, so a converter built while the answer was `null` keeps that `null`:

https://github.com/dahomey-technologies/Dahomey.Cbor/blob/f429ba5/src/Dahomey.Cbor/Serialization/Converters/ObjectConverter.cs#L317

Resolving it lazily instead would cost a dictionary lookup per object read where there is currently a field read — a real trade on the hot path, and the reason this is its own ticket rather than a line in #223.

Three ways out, in increasing cost:

* document that registration must precede the first deserialize with a given `CborOptions`, and leave the behaviour as is;
* make the late registration take effect (registry fix plus lazy or invalidated convention on `ObjectConverter`);
* make the late registration *fail loudly* — throw from `RegisterType` when the base type already carries a resolved `null`, so the no-op stops being silent.

The third is cheap and removes the trap without touching the read path, though it turns a currently harmless call into an exception, which is a breaking change for anyone whose registration is redundant but late.

Happy to implement whichever you prefer.

mcatanzariti:
Fixes #224.

`RegisterType<TSubtype>()` had no effect once anything had asked the registry about the base type, and reported nothing about it. Reading a member declared as the base is enough to get there: that builds the base type's converter, which resolves the convention, and the miss is cached as `null` under the base type. Registering the subtype afterwards was a silent no-op for the life of the `CborOptions`, and the next read failed exactly as it had before the call.

That matters beyond the API doing what its name says. The error a caller meets in this situation is the one #223 is about, and it sends them to register the subtype — advice that, applied to the options already in hand, reproduced the identical error. Whatever wording #223 settles on needs the remedy it names to work.

## What changed

**`DiscriminatorConventionRegistry`** — propagation to base types and interfaces now uses `AddOrUpdate` with an `existing ?? convention` factory instead of `TryAdd`, so it can displace a cached `null` while still leaving a real convention alone. The first hierarchy to claim a supertype keeps it, which is what `TryAdd` already did for that case.

A `Version` counter is incremented when a registration resolves something, and when a convention is added. It exists so a holder of a `null` answer can tell "still nothing" from "worth asking again" without asking.

**`ObjectConverter<T>`** — the registry fix alone is not enough: the converter resolved the convention in its constructor and is itself cached per type, so one built while the answer was `null` kept that `null`. It now re-resolves, but only while the answer is `null` and only when the registry's version has moved since its last attempt.

## On the cost

This touches the read path, so the shape of the check matters. Resolution is monotone — nothing unregisters a type, so a convention once resolved is good forever — which means only a `null` answer can go stale. Steady state is therefore a field read and an int compare, not a dictionary lookup per object read. `ReadingDoesNotMoveTheVersion` pins that reads never move the counter, so the property cannot erode unnoticed.

Two details worth flagging for review:

- The version is sampled **before** the lookup, never after. Sampling after would stamp a registration that landed between the two as already accounted for, and it would never be looked at again.
- The `(convention, version)` pair is written without a lock. The worst a race leaves behind is a `null` with a stale version, which the next read answers by asking once more; two threads resolving concurrently compute the same answer, and once the field is non-null neither writes again.

`DiscriminatorMapping` needed no change — it re-resolves on every call, so the registry fix reaches the write path (`GenerateMemberConverter`, which can throw `Cannot find a discriminator convention`) on its own. That it re-resolves at all turns out to be accidental; filed as #225.

## Tests

8 tests in `Issue0224.cs`: base-class, interface and int-discriminator hierarchies; the base type read directly rather than as a member; a second subtype registered after a successful read; plus two guards — unrelated registrations do not invent a convention, and reads do not move the version.

5 fail on the parent commit. The 2 guards pass either way, which is what a guard should do — they pin behaviour that must not regress rather than the bug.

Full suite green on net8.0 and net10.0 (1970 each, run with `CDDL_REQUIRED=1`), and the generator's 42. net9.0 and netstandard2.0 build but have no local test host; they are on CI.

No README change: the fix removes the ordering constraint rather than documenting it, so there is nothing new for a user to know.

https://claude.ai/code/session_017sFVJR71HtQckyh99DDt4w

---
_Generated by [Claude Code](https://claude.ai/code/session_017sFVJR71HtQckyh99DDt4w)_

### DIFF
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0224.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0224.cs
@@ -197,5 +197,29 @@ namespace Dahomey.Cbor.Tests.Issues
 
             Assert.Null(registry.GetConvention(typeof(Holder)));
         }
+
+        /// <summary>
+        /// The other way into the same trap: a convention registered after a type resolved to nothing.
+        /// </summary>
+        /// <remarks>
+        /// Propagating over a cached null is not enough to cover this one - nothing propagates here,
+        /// because the type asked about is the one carrying the discriminator. The answer has to stop
+        /// being cached at all while it is null, which is what makes the registry consult the newly
+        /// pushed convention instead of repeating what the ones before it said.
+        /// </remarks>
+        [Fact]
+        public void RegisterConventionAfterTheTypeResolvedToNothing()
+        {
+            CborOptions options = new CborOptions();
+            DiscriminatorConventionRegistry registry = options.Registry.DiscriminatorConventionRegistry;
+
+            registry.ClearConventions();
+
+            Assert.Null(registry.GetConvention(typeof(Circle)));
+
+            registry.RegisterConvention(new DefaultDiscriminatorConvention<string>(options.Registry));
+
+            Assert.NotNull(registry.GetConvention(typeof(Circle)));
+        }
     }
 }
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0224.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0224.cs
@@ -1,0 +1,201 @@
+using Dahomey.Cbor.Attributes;
+using Dahomey.Cbor.Serialization.Conventions;
+using Dahomey.Cbor.Tests.Extensions;
+using System;
+using Xunit;
+
+namespace Dahomey.Cbor.Tests.Issues
+{
+    /// <summary>
+    /// Issue #224: registering a subtype had no effect once anything had asked the registry about the
+    /// base type, and said nothing about it. The lookup that resolved nothing cached that null under
+    /// the base type, and the propagation from the subtype could not displace it, so the remedy the
+    /// error message names failed to take on the very options the caller had in hand.
+    /// </summary>
+    /// <remarks>
+    /// Reading a member declared as the base is enough to poison the entry: it builds the base type's
+    /// converter, which resolves the convention. So the sequence below - deserialize, fail, register,
+    /// deserialize again - is the ordinary way into it rather than a contrived one, and every test
+    /// here reuses one <see cref="CborOptions"/> across the registration for that reason.
+    /// </remarks>
+    public class Issue0224
+    {
+        public abstract class Shape { }
+
+        [CborDiscriminator("circle")]
+        public class Circle : Shape { public double Radius { get; set; } }
+
+        [CborDiscriminator("square")]
+        public class Square : Shape { public double Side { get; set; } }
+
+        public class Holder { public Shape Shape { get; set; } }
+
+        public interface IPayload { }
+
+        [CborDiscriminator("text")]
+        public class TextPayload : IPayload { public string Text { get; set; } }
+
+        public class PayloadHolder { public IPayload Payload { get; set; } }
+
+        public abstract class Numbered { }
+
+        [CborIntDiscriminator(7)]
+        public class NumberedSeven : Numbered { public int Value { get; set; } }
+
+        public class NumberedHolder { public Numbered Numbered { get; set; } }
+
+        // a1 655368617065 a2 625f74 66636972636c65 66526164697573 f93e00
+        //   {"Shape": {"_t": "circle", "Radius": 1.5}}
+        private const string CircleInHolder = "A1655368617065A2625F7466636972636C6566526164697573F93E00";
+
+        // {"Payload": {"_t": "text", "Text": "hi"}}
+        private const string TextInHolder = "A1675061796C6F6164A2625F7464746578746454657874626869";
+
+        // {"Numbered": {"_t": 7, "Value": 42}}
+        private const string SevenInHolder = "A1684E756D6265726564A2625F74076556616C7565182A";
+
+        [Fact]
+        public void RegisterTypeAfterTheBaseTypeWasResolved()
+        {
+            CborOptions options = new CborOptions();
+            DiscriminatorConventionRegistry registry = options.Registry.DiscriminatorConventionRegistry;
+
+            Assert.Null(registry.GetConvention(typeof(Shape)));
+
+            registry.RegisterType<Circle>();
+
+            Assert.NotNull(registry.GetConvention(typeof(Shape)));
+        }
+
+        [Fact]
+        public void RegisterTypeAfterAFailedRead()
+        {
+            CborOptions options = new CborOptions();
+
+            // The message is #223's business; what matters here is that the read failed, because that
+            // is what resolved - and used to freeze - the convention for Shape.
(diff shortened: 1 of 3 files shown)


mcatanzariti:
Reviewed the diff and ran the suite locally: green on net8.0, net9.0 and net10.0 (1925 passed, 45 skipped), and the 8 new tests are the whole delta against `master`. The diagnosis holds and both locks on the bug are lifted. Two things I would change before merging, plus two notes.

### 1. A convention registered after a read still does not take effect

`RegisterConvention` increments the version, but nothing can act on it. When a holder takes the hint and asks again, the ask lands on `GetConvention`, which hands back the very `null` it cached before the new convention was pushed. That is the same bug as #224, entered through the other door — it is out of scope for this PR, but the increment as it stands suggests the case is handled when it is not.

Rather than sweeping the cached nulls on registration (which races with `Propagate`: an entry that turned non-null between the read and the removal would be dropped and then re-resolve to `null` for a supertype that carries no discriminator of its own), the smaller change is to stop caching nulls at all:

```csharp
public IDiscriminatorConvention? GetConvention(Type type)
{
    if (_conventionsByType.TryGetValue(type, out IDiscriminatorConvention? convention))
    {
        return convention;
    }

    convention = InternalGetConvention(type);

    if (convention != null)
    {
        _conventionsByType.TryAdd(type, convention);
    }

    return convention;
}
```

The dictionary then holds only resolutions that found something, and the staleness this PR works around stops existing: `Propagate` can go back to a plain `TryAdd`, since the entries that were in its way were exactly those nulls. What it costs is a re-resolution per lookup of a type that has no discriminator — `TryRegisterType` is an `ObjectMappingRegistry.Lookup` plus a null check, both cached — and the callers that ask often, `ObjectConverter<T>` first among them, already hold their own answer and only come back when `Version` has moved. So the version counter is still needed; it just stops being the only thing standing between a registration and its effect.

A test that fails on this PR's head and passes with the above:

```csharp
[Fact]
public void RegisterConventionAfterTheTypeResolvedToNothing()
{
    CborOptions options = new CborOptions();
    DiscriminatorConventionRegistry registry = options.Registry.DiscriminatorConventionRegistry;

    registry.ClearConventions();

    Assert.Null(registry.GetConvention(typeof(Circle)));

    registry.RegisterConvention(new DefaultDiscriminatorConvention<string>(options.Registry));

    Assert.NotNull(registry.GetConvention(typeof(Circle)));
}
```

Note that propagating over a cached null does not cover this one: nothing propagates here, because the type asked about is the one carrying the discriminator.

### 2. The convention/version pair needs a barrier

`_discriminatorConvention` is written before `_discriminatorConventionVersion`, and `GetDiscriminatorConvention` reads them in the same order. The remark on the field justifies the unsynchronized write by "a torn write costs nothing worse than resolving again", which is true of the interleaving but not of the reordering: on a weak memory model (arm64) a reader can see the new version next to the `null` it replaced, take that for a current answer, and return `null` — a read failing with the #223 error while the convention is registered. It heals on the next call, but the caller has already seen the throw.

Publishing the version last and reading it first closes it, for nothing measurable:

```csharp
// in GetDiscriminatorConvention
int resolvedVersion = Volatile.Read(ref _discriminatorConventionVersion);
IDiscriminatorConvention? convention = _discriminatorConvention;

// in ResolveDiscriminatorConvention
_discriminatorConvention = convention;
Volatile.Write(ref _discriminatorConventionVersion, version);
```

### 3. `ClearConventions` leaves the cache and the version untouched

Pre-existing, and harmless while nothing depends on the version being exact. Worth a line saying the cache deliberately outlives the conventions, now that `Version` reads as a contract.

### 4. `_discriminatorConvention!` at the member-name comparison

That site still reads the field directly rather than through `GetDiscriminatorConvention`. It is safe — the `_hasDiscriminatorMember` guard implies this type's own mapping carries a discriminator, so the constructor's resolution was non-null — but it is now the only direct read left, and a short comment would keep it from looking like an oversight.

---

Nothing above touches the correctness of the fix itself: the `AddOrUpdate` propagation and the version gate do what they claim, and `ReadingDoesNotMoveTheVersion` is a good thing to have pinned. With 1 and 2 applied the suite stays green (1926 passed across the three TFMs).

PR #228, opened 2026-08-15
Name the registration a discriminated document needs, closes #223

### LINKED ISSUE #223: Unregistered subtype reports a missing CreatorMapping rather than the registration the README prescribes

Rafael-SOWNet:
Closes #223. Built to your scoping — three formats, current message kept where no discriminator is present, `CborValue` for the value, `MemberMappingErrors` for the wording.

## Where it goes

Your diagnosis was exactly right and it decided the placement. `ObjectConverter<Shape>` resolves its convention in the constructor, `Shape` carries no discriminator of its own, so `_discriminatorConvention` stays null and the read falls to the `else` at `ObjectConverter.cs:734` where `CreateInstance()` runs before any member is read. The document is now asked, there, whether it carries a discriminator after all:

```
no type is registered for the discriminator circle carried by this document, so it cannot be read
as Shape; call CborOptions.Registry.DiscriminatorConventionRegistry.RegisterType<T>() for the
concrete type that discriminator denotes
```

The value is named as well as the remedy, because the remedy is per subtype — a caller reading a hierarchy registers each concrete type, and the value is what says which one this document needed.

## The three arms

Probed where each format puts it, mirroring the resolution arms above and taking a bookmark the same way: `FindItem(convention.MemberName)`, `FindItem(0)`, and `TryReadSemanticTag` against `DiscriminatorSemanticTag`.

**Every registered convention is asked**, not just a default, because which one would have written the key is precisely what is unknown here — the type resolved to none. `DiscriminatorConventionRegistry.Conventions` is the `internal` accessor you predicted I would want; it exposes registration order, which is why it is not public.

**The value is read as a `CborValue`.** Going through the convention would resolve it to a `Type` and fail on the very value being reported, and this way one line renders a string and an integer discriminator without knowing which it is — `"circle"` and `7` both come out.

## What does not change

Where no convention finds anything, the document really is undiscriminated and the missing `CreatorMapping` is the correct answer. Two tests pin that: `AnUndiscriminatedDocumentStillReportsTheMissingCreatorMapping`, and `WithNoConventionsRegisteredTheOriginalMessageStands` — the second also pins that the probe asks the registry rather than assuming `"_t"`.

## On #226

`TheRemedyTheMessageNamesResolvesTheDocument` registers on a fresh `CborOptions` before any read, so it passes with or without your fix. But the point you made stands and is the reason this message is worth having at all: a reader who hits it, adds the call to the options **already in hand** and retries needs #226 to get a `Circle` back. Without it they would get the same error twice and conclude the advice was wrong. Merging this ahead of #226 would ship advice that is only conditionally true, so I would rather it went in after — no rebase needed either way, the files do not overlap as you said.

I dropped the README line, per your correction.

## Tests

Seven in `Issues/Issue0223.cs`: the default format naming both remedy and value, an interface rather than an abstract class, `IntKeyMap` with an integer discriminator, `Array` behind the semantic tag, the remedy actually resolving the document, and the two negative controls above.

They write on an options instance of their own, deliberately. Writing registers the runtime type as a side effect of building its converter, so sharing one instance with the read — or going through `CborOptions.Default`, as `Helper.Write` does — registers the very subtype the test is about. The interface case passed spuriously until I separated them, purely on test-execution order.

**1969 library tests and 42 generator tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped, four TFMs building. **Four of the seven fail without the `src/Dahomey.Cbor/` change**; the other three are the negative controls and pass either way, which is what they are for.

---

### DIFF
--- a/src/Dahomey.Cbor.Tests/Issues/Issue0223.cs
+++ b/src/Dahomey.Cbor.Tests/Issues/Issue0223.cs
@@ -45,7 +45,7 @@ namespace Dahomey.Cbor.Tests.Issues
                 () => Cbor.Deserialize<Shape>(hex.HexToBytes(), new CborOptions()));
 
             Assert.Contains(ExpectedRemedy, exception.Message);
-            Assert.Contains("circle", exception.Message);
+            Assert.Contains("discriminator \"circle\"", exception.Message);
             Assert.DoesNotContain("CreatorMapping", exception.Message);
         }
 
@@ -59,7 +59,7 @@ namespace Dahomey.Cbor.Tests.Issues
                 () => Cbor.Deserialize<IShape>(hex.HexToBytes(), new CborOptions()));
 
             Assert.Contains(ExpectedRemedy, exception.Message);
-            Assert.Contains("circle", exception.Message);
+            Assert.Contains("discriminator \"circle\"", exception.Message);
         }
 
         /// <summary>
@@ -76,7 +76,7 @@ namespace Dahomey.Cbor.Tests.Issues
                 () => Cbor.Deserialize<IntShape>(hex.HexToBytes(), new CborOptions()));
 
             Assert.Contains(ExpectedRemedy, exception.Message);
-            Assert.Contains("7", exception.Message);
+            Assert.Contains("discriminator 7", exception.Message);
         }
 
         /// <summary>
@@ -86,13 +86,13 @@ namespace Dahomey.Cbor.Tests.Issues
         [Fact]
         public void AnUnregisteredSubtypeUnderArrayFormatIsReported()
         {
-            string hex = WriteOnItsOwnOptions<ArrayShape>(new ArrayCircle { Radius = 1.5 });
+            string hex = WriteOnItsOwnOptions<ArrayShape>(new ArraySquare { Radius = 1.5 });
 
             CborException exception = Assert.Throws<CborException>(
                 () => Cbor.Deserialize<ArrayShape>(hex.HexToBytes(), new CborOptions()));
 
             Assert.Contains(ExpectedRemedy, exception.Message);
-            Assert.Contains("square", exception.Message);
+            Assert.Contains("discriminator \"square\"", exception.Message);
         }
 
         /// <summary>
@@ -129,9 +129,9 @@ namespace Dahomey.Cbor.Tests.Issues
         }
 
         /// <summary>
-        /// Clearing the conventions leaves nothing to probe with, so the document cannot be interrogated
-        /// and the original message stands. Pins that the probe asks the registry rather than assuming
-        /// the default member name.
+        /// Under the default format the member name is the convention's to choose, so clearing the
+        /// conventions leaves nothing to probe with and the original message stands. Pins that the probe
+        /// asks the registry rather than assuming the default member name.
         /// </summary>
         [Fact]
         public void WithNoConventionsRegisteredTheOriginalMessageStands()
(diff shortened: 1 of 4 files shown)


Rafael-SOWNet:
Force-pushed to `2f9b78d`: the first push swept in an unrelated test file of mine (`RefusalReachesEveryEntryPointTests.cs`, a contract theory for #219 that has been waiting on #220). It is gone from this branch and will come as its own PR. The count in the description is corrected with it — **1969**, not 1981; the difference was that file's twelve cases.

Nothing else changed.

mcatanzariti:
Reviewed, and pushed two changes onto the branch — a rebase onto `master`, and one commit (`683576a`) on the points below. All of it is on your work; the diagnosis and the placement were right.

**Rebased.** The branch was cut from `f429ba5`, before both #226 and #225 landed — and those touch `DiscriminatorConventionRegistry.cs` and `ObjectConverter.cs`, the two files this changes. No conflicts, but the overlap was real, and it matters for the reason you gave yourself: with #226 in, the remedy the message names now works on the options the caller already holds, so the advice is unconditionally true.

**The convention loop only belongs to one arm.** `ThrowIfADiscriminatorIsPresent` looped over every registered convention in all three formats, but only `StringKeyMap` ever reads one — `IntKeyMap` looks at index 0 and `Array` at the semantic tag, both fixed by the format, not chosen by a convention. So those two re-ran an identical probe once per registered convention, and, less harmlessly, `ClearConventions()` silenced them even though no convention had a say in where they looked. Each format is now probed on its own terms, and `WithNoConventionsRegisteredAPositionalDiscriminatorIsStillFound` pins that the positional two still answer with the registry emptied — sitting next to your `WithNoConventionsRegisteredTheOriginalMessageStands`, which stays exactly as it was, since for the named-member format clearing really does leave nothing to look for.

**What the probe cannot tell apart, written down.** A type that is not polymorphic but declares a member where a discriminator would sit — `_t`, or index 0 — is now reported as an unregistered subtype rather than as the missing creator it is. That is not a regression: the resolution path above reads the same place on the same evidence, and the document offers nothing else to go on. It is recorded in the remarks so the next reader does not have to rediscover it.

**Assertions.** `Assert.Contains("7", ...)` passes on any message containing a 7 anywhere — a type name, another number. The value assertions now match the value inside the phrase that carries it. Worth noting while tightening them: a string discriminator renders quoted, so the message reads `the discriminator "circle"`, not `circle` as the description says. Better that way, I think — it separates the value from the prose around it.

**`ArrayCircle` carried the discriminator `"square"`.** Renamed to `ArraySquare`.

**Tests:** 1937 passing on net8.0, net9.0 and net10.0, plus 42 generator tests. The 45 CDDL tests are skipped here — the `cddl` gem is not installed on this machine, unrelated to the change — so `CddlPolymorphismTests` is on CI to confirm.

Merging now.

PR #229, opened 2026-08-15
Assert a refusal reaches the caller through every read entry point

Rafael-SOWNet:
Test-only, and a follow-up to your #220 rather than a fix for anything still broken.

## What it is for

#219 survived as long as it did because the contract held at every public entry point **except one**, and the tests were grouped *by mechanism* rather than *by entry point* — so nothing ever compared the doors against each other. Each door had its own file, each file tested what it was about, and the odd one out was only visible from a vantage point no test occupied.

`Issues/Issue0219.cs` fixes the case. This closes the shape: one refusing converter, one whole document, the same assertion on each of

```
Deserialize(ReadOnlySpan)          DeserializeAsync(Stream)
Deserialize(ReadOnlySequence)      DeserializeAsync(PipeReader)
DeserializeMultiple(ReadOnlySpan)  ReadNextItemAsync(PipeReader)
DeserializeMultiple(ReadOnlySequence)
```

twelve cases with the `[CborConstructor]` half.

## Two details that are the point

**A theory over the entry points, not a test each.** A failure names the door rather than the mechanism, and adding an entry point without a row is a visible omission rather than a silent one.

**It asserts against the substitute messages by name**, not just the exception type:

```csharp
Assert.StartsWith(ConverterMessage, exception.Message);
Assert.DoesNotContain("no item was read", exception.Message);
Assert.DoesNotContain("Unexpected end of buffer", exception.Message);
```

A `CborException` of the right type passes while the message has been replaced, and a replaced message is exactly how a refusal disguises itself as malformed input — which is the whole of #219.

## Verification

**1974 library tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped, four TFMs building.

All twelve pass on `master` as of #220. Against `f429ba5^` — the commit before it — exactly two fail, and they are the two `ReadNextItemAsync` rows, both reporting `Reader has completed with some buffer bytes but no item was read`. So it isolates your fix precisely.

Offered on #220 before it merged; opening it separately now, as you'd expect.

---

### DIFF
--- a/src/Dahomey.Cbor.Tests/RefusalReachesEveryEntryPointTests.cs
+++ b/src/Dahomey.Cbor.Tests/RefusalReachesEveryEntryPointTests.cs
@@ -26,7 +26,10 @@ namespace Dahomey.Cbor.Tests
     /// <para>
     /// Arranged as a theory over the entry points rather than as a test each, so the failure names the
     /// door rather than the mechanism, and so adding an entry point without a row here is a visible
-    /// omission rather than a silent one.
+    /// omission rather than a silent one. The rows are the reading half of <see cref="Cbor"/>: the four
+    /// single-item doors, the four multiple-item ones, and <c>ReadNextItemAsync</c>. The non-generic
+    /// <c>Type</c> overloads are the same methods with the type passed rather than inferred, and reach
+    /// the converter by the same route, so they are not repeated here.
     /// </para>
     /// </remarks>
     public class RefusalReachesEveryEntryPointTests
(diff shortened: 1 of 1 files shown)


mcatanzariti:
Rebased on `master` (#228 is in) and pushed two changes on top of yours.

**The theory was one crossing short of its own claim.** It named seven doors; `DeserializeMultipleAsync(Stream)` and `DeserializeMultipleAsync(PipeReader)` were not among them — which is precisely "many items" × "a stream that may not have them all yet", the crossing #219 lived at. The creator theory was two rows shorter still, with no reason for the difference. A theory whose argument is that a missing row is a visible omission cannot itself have one, so both refusing types now run the same nine rows, generated once by `EntryPointsFor<T>` rather than listed twice. Twelve cases became eighteen. The non-generic `Type` overloads are deliberately not repeated — same methods, type passed rather than inferred — and the remark says so.

**The delegates are awaited.** `Func<..., Task<object>>` and `async Task` theories, so no `GetAwaiter().GetResult()`: six `VSTHRD002` warnings gone. The synchronous doors wrap their result instead, since a lambda holding a `ReadOnlySpan` cannot be `async`; the exception still surfaces through `Assert.ThrowsAsync`, which catches a delegate that throws while being invoked. And `name` is now used rather than unread — it carries the failure message on the primary assertion, which is what the column was for and also clears `xUnit1026`. The file adds no build warnings now.

**Verified here.** 2000 tests green on net8.0, net9.0 and net10.0 (45 CDDL cases skipped locally — the `cddl` tool isn't installed on this machine). Your isolation claim reproduces after the change: against `f429ba5^`, exactly two of the eighteen fail, both `ReadNextItemAsync` rows, on `Reader has completed with some buffer bytes but no item was read`.

The framing was right — grouping by entry point rather than by mechanism is what makes the odd door visible, and asserting against the substitute message by name is what makes the test worth having. Thanks. Merging shortly.

PR #232, opened 2026-08-16
Resolve an Array member by its position, closes #222

### LINKED ISSUE #222: CborObjectFormat.Array on a type with no discriminator loses its members unless the indexes start at 0
`CborObjectFormat.Array` writes its members packed from position 0 in index order, while the read maps position to `[CborProperty(n)]` index literally. The two only agree when the indexes happen to be 0-based and contiguous, and a type carrying no discriminator has no reason to declare them that way.

Measured on `de6acd3`, plain `new CborOptions()`, no generated context involved. Same two members, same values, only the declared indexes differ:

```csharp
[CborObjectFormat(CborObjectFormat.Array)]
public class Row
{
    [CborProperty(A)] public int Id { get; set; }
    [CborProperty(B)] public string Name { get; set; }
}

Helper.Write(new Row { Id = 2, Name = "row" });     // 82 02 63726F77  -- [2, "row"] in every case
Cbor.Deserialize<Row>(bytes);
```

| indexes | written | read back |
|---|---|---|
| `0`, `1` | `82 02 63726F77` | correct |
| `1`, `2` | `82 02 63726F77` | **throws** `[3] Invalid major type TextString` at `$.Id` |
| `5`, `9` | `82 02 63726F77` | **`Id = 0`, `Name = ""`** — no exception, both members lost |

The third row is the one that worries me: valid-looking bytes, no error, and every member silently defaulted. A caller writing a document and reading it back in the same process gets an object with nothing in it.

## Why the discriminated case does not show this

Add a discriminator to the same type and indexes `1`, `2` round-trip correctly:

```csharp
[CborObjectFormat(CborObjectFormat.Array)]
[CborDiscriminator("d")]
public class Row { [CborProperty(1)] public int Id { get; set; } [CborProperty(2)] public string Name { get; set; } }
// writes 82 02 63726F77 under the default Auto policy, and reads back Id=2 Name=row
```

`ObjectConverter`'s `case CborObjectFormat.Array:` read block runs only when `DiscriminatorConventionRegistry.GetConvention` returns a convention for the type being read, and it returns null for a type carrying no discriminator — so an undiscriminated `Array` type is not read through the `Array` path at all. That is also why the documented layout in the README (`Array: [39(2), 1, 3.0]`, discriminator first, then index 1, then index 2) describes only half the cases.

## The decision this needs

Both halves are defensible and it is your call which one moves:

* **The reader is right** — `[CborProperty(n)]` means position *n*, always, so position 0 stays reserved whether or not a discriminator is written. Costs one item per object for undiscriminated types, and keeps indexes stable if a discriminator is added later.
* **The writer is right** — an undiscriminated type has no slot 0 to reserve, so packing from 0 is the compact and obvious encoding. Then adding `[CborDiscriminator]` to an existing type shifts every position by one and silently changes the wire format of every document already written.

I lean to the first: a type gaining a discriminator later is the more likely edit, and it should not reinterpret existing documents.

Either way there is no compatibility cost to fix it, because the current output for the affected shapes cannot be read back by this library at all — so no round-trippable document exists in that form today.

Happy to open a PR for whichever you prefer. Found while enrolling the CDDL fixtures in `GeneratedCorpusTests`; `CddlRow` sits on that file's `NotReadBackByEitherPath` exclusion list because of this.

mcatanzariti:
Fixes #222.

`CborObjectFormat.Array` writes its members packed in ascending index order and emits no keys, so a position is all the document supplies. The read counted positions and used the count as the declared member index — the same number only for a type whose indexes start at zero and run consecutively. Every other type lost its members: indexes `1`, `2` threw `Invalid major type TextString` reading the name into the id, and indexes far enough out returned a fully defaulted object with no error at all.

## The index orders, it does not address

`ObjectMapping` sorts by the index and allows gaps and negatives alike, and a negative index is already pinned by `DeterministicEncodingTests.ArrayFormatWithNegativeIndex` — so an index cannot be a position. The read now resolves a position against the member list and keys everything downstream (read lookups, `IsCreatorMember`, the `creatorValuesByIndex`/`regularValuesByIndex` dictionaries, `PushMemberSegment`) on the *declared* index it finds there — the same key those structures were built from, which is why the creator and struct paths need no change of their own.

## Two consequences beyond the reported shapes

- **The discriminated case was affected too**, whenever the members did not start at 1: the read reserved exactly one slot and assumed they carried on from there.
- **That slot was reserved on registry state.** The `context.memberIndex++` ran whenever a discriminator convention merely *resolved* for the type, which the registry answers for every base of a registered hierarchy — so registering a derived type elsewhere in the program changed how the base read its own bytes while leaving the write alone. Positions are now counted over members only, and the discriminator is recognised by its semantic tag as it already was.

## Compatibility and docs

No writer change and no format change: the affected shapes could not be read back by this library at all, so no round-trippable document existed in that form. `CddlRow`, `CddlArrayBase` and `CddlArrayDerived` come off `GeneratedCorpusTests`' `NotReadBackByEitherPath` exclusion list and round-trip there now.

Contiguity is deliberately not validated — under an ordering reading a gap is not a mistake, and a rule would have to reject the legitimate negative-index declaration. Documented instead in the README and on `SetObjectFormat`: in `Array` the index orders members and gaps never reach the wire, so two types differing only in index values encode identically; `IntKeyMap` is the format whose index survives as an address.

## Verification

`net8.0`, `net9.0` and `net10.0`: 1967 passed, 0 failed, plus 42 generator tests. The 45 CDDL tests skip here because the external `cddl` tool is not installed on this machine — they need a run where it is present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,28 @@ If the aim is compactness rather than interoperability with a producer you do no
 keys altogether. Both compress a document of repeated keys harder than string references do, in plain
 RFC 8949.
 
+The two read `[CborProperty(n)]` differently, and the difference is worth being explicit about.
+`IntKeyMap` writes `n` into the document as the member's key, so it addresses the member: gaps are
+free, and a member keeps its meaning wherever it moves in the type. `Array` writes no keys at all, so
+`n` only *orders* the members — they are written in ascending index order and read back by position.
+Gaps and negative indexes are allowed and change nothing on the wire, so these two types produce
+identical bytes:
+
+```csharp
+[CborObjectFormat(CborObjectFormat.Array)]
+public class Row { [CborProperty(0)] public int Id { get; set; } [CborProperty(1)] public string Name { get; set; } }
+
+[CborObjectFormat(CborObjectFormat.Array)]
+public class SameRow { [CborProperty(5)] public int Id { get; set; } [CborProperty(9)] public string Name { get; set; } }
+
+// both write 82 02 63 726F77  --  [2, "row"]
+```
+
+What that costs is that an `Array` type's wire format depends on its member *set*, not only on the
+indexes: inserting a member with an index that sorts between two existing ones shifts everything after
+it by one position, where `IntKeyMap` would not move. Use `IntKeyMap` where indexes need to be stable
+addresses across versions.
+
 ### Bignums (RFC 8949 §3.4.3)
 
 A member typed `System.Numerics.BigInteger` reads and writes integers of any width. Values that fit in 64
--- a/src/Dahomey.Cbor.Tests/GeneratedCorpusTests.cs
+++ b/src/Dahomey.Cbor.Tests/GeneratedCorpusTests.cs
@@ -615,20 +615,21 @@ namespace Dahomey.Cbor.Tests
         /// with a reason attached and not something a shape can drift into.
         /// </summary>
         /// <remarks>
-        /// Both entries were reproduced against a plain <c>new CborOptions()</c> with no generated
-        /// context in play, so they are pre-existing library defects rather than anything source
-        /// generation does differently. All of them stay in
-        /// <see cref="GeneratedBytesMatchReflectionBytes"/>, which they pass: the two writers agree
+        /// The entry was reproduced against a plain <c>new CborOptions()</c> with no generated
+        /// context in play, so it is a pre-existing library defect rather than anything source
+        /// generation does differently. It stays in
+        /// <see cref="GeneratedBytesMatchReflectionBytes"/>, which it passes: the two writers agree
         /// byte for byte and it is the shared reader that cannot consume the result.
         /// <list type="bullet">
         /// <item><description>A member declared as an abstract class or an interface. The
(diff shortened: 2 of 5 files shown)


mcatanzariti:
Follow-up filed as #233: `[CborIgnoreIfDefault]` (and `ShouldSerialize` generally) still shifts every position after the omitted member in `Array` format. Verified to fail identically on `master` before this PR, byte for byte, so it is untouched by this change rather than introduced by it — the fix belongs in the writer, which has to hold the slot a keyless format cannot express as absent.

mcatanzariti:
Rerun with the `cddl` gem installed locally, so the CDDL tests execute rather than skip: **2012 passed, 0 failed, 0 skipped** on net8.0, net9.0 and net10.0 with `CDDL_REQUIRED=1`, plus the 42 generator tests. That closes the one gap left in the verification note above — the schema-validation coverage over `CddlRow`, `CddlArrayBase` and `CddlArrayDerived` is now confirmed green here and not only expected to be so in CI.

PR #235, opened 2026-08-16
Write every Array position, closes #233

### LINKED ISSUE #233: CborObjectFormat.Array: a member skipped by ShouldSerialize shifts every position after it
`CborObjectFormat.Array` writes members positionally and emits no keys, so every member must hold its position for the document to be readable. `ShouldSerialize` is applied in that format all the same: a member it declines is simply not written, and everything after it slides one position earlier. The read then resolves each position against the member list and lands on the wrong member.

Measured on the head of #232, plain `new CborOptions()`, no generated context involved:

```csharp
[CborObjectFormat(CborObjectFormat.Array)]
public class Skipper
{
    [CborProperty(5)] public int Id { get; set; }
    [CborProperty(7)][CborIgnoreIfDefault] public string? Name { get; set; }
    [CborProperty(9)] public int Tail { get; set; }
}

Helper.Write(new Skipper { Id = 1, Name = null, Tail = 42 });  // 82 01 182A -- [1, 42]
Helper.Read<Skipper>("8201182A");
// throws [3] Expected major type TextString (3). Failed to deserialize from "$.Name"
```

`Tail`'s value arrives at `Name`'s position. Here the two types disagree and it throws; where the shifted members share a type it will not, and the values land silently on the wrong members — the same failure mode as the third row of #222.

## Not a regression

The same repro fails identically on `master` before #232, byte for byte: the old read counted a position too, so a hole shifted it just as much. #232 changed which member a position resolves to, not whether a hole can exist, so this is untouched by it either way and wants its own fix.

## Where the fix belongs

The writer, not the reader. A reader cannot recover a position the document never carried — an omitted member leaves no trace to skip over. `Array` has no way to express absence, so a member that holds a position has to be written whatever `ShouldSerialize` says, presumably as `null` (or the type's default) to keep the slot. That is the same reasoning that already excludes `Array` from the deterministic member sort in `MemberConvertersForWrite`: reordering or removing members in a keyless format changes what the document means, not merely how it is spelled.

`GetSize` needs to follow, since it counts the same predicate to size the definite-length array.

Worth deciding at the same time whether `[CborIgnoreIfDefault]` on an `Array` type should instead be refused at registration in `ObjectMapping.ValidateMemberNamesAndIndexes`, alongside the existing checks that every member carries an index and that no two share one. Refusing it is louder and costs a declaration that has no meaning in this format; honouring it by writing the slot keeps the attribute harmless where a type uses one format for some documents and another for others.

mcatanzariti:
Closes #233.

`CborObjectFormat.Array` writes its members positionally and emits no keys, so a member is identified
by nothing but its position. `ShouldSerialize` was applied there all the same: a member it declined
was not written and everything after it slid one position earlier, so the read resolved each position
against the member list and landed on the wrong member — throwing where the shifted members disagreed
on type, assigning silently where they agreed.

```csharp
[CborObjectFormat(CborObjectFormat.Array)]
public class Skipper
{
    [CborProperty(5)] public int Id { get; set; }
    [CborProperty(7)][CborIgnoreIfDefault] public string? Name { get; set; }
    [CborProperty(9)] public int Tail { get; set; }
}

// before: 82 01 182A  --  [1, 42], and the read throws on Tail's value arriving at Name
// after:  83 01 F6 182A  --  [1, null, 42]
```

## The fix is the writer's

A reader cannot recover a position the document never carried — an omitted member leaves no trace to
skip over — and the format has no way to express absence. So a member that holds a position is written
whatever the predicate says, with the value it holds: `null` or the type's default in the very case
`[CborIgnoreIfDefault]`, a `DefaultValue` or a `ShouldSerializeXyz()` method exists to catch. `GetSize`
counts the same way, since it sizes the definite-length array from the same predicate. That is the
reasoning that already keeps `Array` out of the deterministic member sort: removing or reordering
members in a keyless format changes what the document means, not how it is spelled.

The discriminator is the exception and keeps answering for itself. It is not a member of the type and
holds no position, being written under a semantic tag and recognised by it, so whether it appears at
all remains the discriminator policy's business — `WriteArrayWithDiscrimator` and
`Issue0222.DiscriminatedIndexesNeedNotStartAtOne` pin that.

The two map formats are untouched: a key travels with each value there to say which member it belongs
to, so an omitted member costs nothing and the predicate keeps working as declared.

## Decisions

The issue asked whether `[CborIgnoreIfDefault]` on an `Array` type should instead be refused at
registration. This honours it by writing the slot rather than refusing it, which keeps the attribute
harmless on a type that uses one format for some documents and another for others.

Struct coverage was left out because `Array` + struct is unreachable today: `StructMemberConverter`
lacks the null-`MemberName` guard `MemberConverter` has, so any struct declaring `[CborProperty(int)]`
throws `ArgumentNullException` when its converter is built. Filed separately.

## Tests

`Issue0233` covers the reported shape, its round trip, the silent shape via `ShouldSerializeXyz()` on
three members of one type, a present value as a non-regression, and the map still omitting. Full suite
green with `CDDL_REQUIRED=1`: 2017 passing, 0 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,14 @@ indexes: inserting a member with an index that sorts between two existing ones s
 it by one position, where `IntKeyMap` would not move. Use `IntKeyMap` where indexes need to be stable
 addresses across versions.
 
+For the same reason, `Array` writes every member of the type on every document. A member omitted by
+`[CborIgnoreIfDefault]`, `[CborProperty(DefaultValue = ...)]` or a `ShouldSerializeXyz()` method would
+leave no trace in a keyless format, so everything after it would shift a position earlier and read
+back onto the wrong member. Those declarations are therefore ignored in `Array`, and the member is
+written with the value it holds — `null` or the type's default in the case they exist to catch. They
+work as declared in `StringKeyMap` and `IntKeyMap`, where each value travels with the key that says
+which member it belongs to.
+
 ### Bignums (RFC 8949 §3.4.3)
 
 A member typed `System.Numerics.BigInteger` reads and writes integers of any width. Values that fit in 64
--- a/README.md
+++ b/README.md
@@ -296,12 +296,26 @@ it by one position, where `IntKeyMap` would not move. Use `IntKeyMap` where inde
 addresses across versions.
 
 For the same reason, `Array` writes every member of the type on every document. A member omitted by
-`[CborIgnoreIfDefault]`, `[CborProperty(DefaultValue = ...)]` or a `ShouldSerializeXyz()` method would
-leave no trace in a keyless format, so everything after it would shift a position earlier and read
-back onto the wrong member. Those declarations are therefore ignored in `Array`, and the member is
-written with the value it holds — `null` or the type's default in the case they exist to catch. They
-work as declared in `StringKeyMap` and `IntKeyMap`, where each value travels with the key that says
-which member it belongs to.
+`[CborIgnoreIfDefault]` or by a `ShouldSerializeXyz()` method would leave no trace in a keyless
+format, so everything after it would shift a position earlier and read back onto the wrong member.
+Those declarations are therefore ignored in `Array`, and the member is written with the value it
+holds — `null` or the type's default in the case they exist to catch. They work as declared in
+`StringKeyMap` and `IntKeyMap`, where each value travels with the key that says which member it
+belongs to.
+
+One combination becomes visible that way: a member declaring both `[CborIgnoreIfDefault]` and
+`[CborRequired]` — with a policy of `DisallowNull` or `Always` — now throws when it is null, where the
+omission used to hide it. The two declarations contradict each other in a format that cannot omit
+anything, and the document the write used to produce could not be read back. Drop one of the two, or
+use `IntKeyMap`.
+
+A member keyed by an index has no name to put in such a message, so it is reported as its index:
+`Property 'index 2' cannot be null.` That also applies to `IntKeyMap`, where the same messages used to
+name the member with an empty string.
+
+The discriminator is not a member and holds no position — it is written under a semantic tag and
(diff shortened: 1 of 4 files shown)


mcatanzariti:
A review pass over the first commit found a real defect in it, fixed by `4fa4809`.

**The discriminator exclusion was type-dependent.** `HoldsAPosition` tested for
`DiscriminatorMemberConverter<T>`, with `T` the *declared* type. A write resolves `memberConvertersForWrite`
on the object's *actual* type, so a polymorphic write hands a declared type's converter the derived type's
members — including a discriminator converter closed over that derived type, which the test did not
recognise. The discriminator was then written unconditionally, in the only case where one is written at all:
with `CborDiscriminatorPolicy.Never` over an `Array` hierarchy the write emitted it regardless. No existing
test caught it, because under `Auto` the policy wants it there and the two behaviours coincide.

It is now marked by a non-generic `IDiscriminatorMemberConverter` and recognised through that.
`TheDiscriminatorPolicyStillDecidesInAnArray` pins it and fails on the first commit.

**Two consequences documented at the same time.**

A member declaring both `[CborIgnoreIfDefault]` and `[CborRequired]` (`DisallowNull`/`Always`) now reports
the null it used to hide behind a document that could not be read back — the pair contradicts itself in a
format that cannot omit anything. That message named the member with an empty string whenever it carried an
index rather than a name, in `IntKeyMap` as much as in `Array`, so `MemberConverter` falls back to naming it
by its index: `Property 'index 2' cannot be null.`

The README also lost a mention of a `DefaultValue` on `CborPropertyAttribute`, which does not exist — what
omits a member is `[CborIgnoreIfDefault]`, with `System.ComponentModel`'s `[DefaultValue]` deciding only what
"default" means, or a `ShouldSerializeXyz()` method.

Full suite green on net8.0/9.0/10.0 with `CDDL_REQUIRED=1`: 2019 passing, 0 skipped.

PR #238, opened 2026-08-16
Generate the collection interfaces instead of refusing them

Rafael-SOWNet:
A generated context refused a model that declares `IList<T>` rather than `List<T>`:

```
CBOR1002: 'System.Collections.Generic.IList<int>' cannot be handled by CBOR source generation
(collection interfaces and abstract collection types are not generated yet; use a concrete type
such as List<T>). It would fall back to runtime reflection, which fails under Native AOT
```

Declaring the interface is expressing a contract, not being awkward, and the remedy the message offers is to change the model. This closes it for the six interfaces the reflection path already resolves.

## Why they were refused

Nothing exotic: `IList<T>` **implements** `ICollection<T>`, so it reached the concrete-collection branch in `TypeCollector.Classify`, and was then turned away by the parameterless-constructor test. The interfaces are now classified before those tests.

## What they resolve to

Exactly what `CollectionConverterProvider` resolves them to, because a document written through one path has to read back through the other:

| Declared | Converter |
|---|---|
| `IList<T>`, `ICollection<T>`, `IEnumerable<T>`, `IReadOnlyList<T>`, `IReadOnlyCollection<T>` | `InterfaceCollectionConverter<List<T>, I, T>` |
| `ISet<T>` | `InterfaceCollectionConverter<HashSet<T>, ISet<T>, T>` |
| `IDictionary<K,V>` | `InterfaceDictionaryConverter<K,V>` |

**The backing type is the whole reason these could not be generated before.** An interface cannot be instantiated, so the read fills a concrete collection and hands it back as the interface — and that concrete type is what the reflection path reaches `MakeGenericType` for. Nothing in the source mentions `List<T>` merely because a member said `IList<T>`, so `TypeModel` carries it as a string: there is no symbol to reach for, and the emitter only ever needs the name.

`IReadOnlyList<T>` and `IReadOnlyCollection<T>` are included even though a caller cannot add to what they get back — the converter builds a `List<T>` and returns it as the read-only interface, which is what the reflection path does.

## CDDL

Rendered as the document rather than as the backing type — `[* X]` and `{* K => V}`, the same as the concrete forms. Which concrete type the reader hands back is not something the wire records.

## Tests

- **Seven generator cases**, one per interface, asserting no diagnostic. All seven report `CBOR1002` without the change.
- **A corpus entry.** `GeneratedInterfaceHolder` is enrolled in `GeneratedCorpusTests`, which is where byte identity against the reflection path is actually asserted, along with the round trip.
- **One thing the corpus cannot see**: that `ISet<T>` comes back a `HashSet<T>` and not a `List<T>`. Both write identical bytes, so a byte comparison is blind to it — and getting it wrong would hand the caller something that does not satisfy the interface they declared.
- **The emitted schema**, pinned. `EveryEmittedSchemaParses` already runs it through the reference `cddl` tool by assembly scan, which says it is well-formed; this says it is right, since a rendering that lost the element type would parse just as happily.

## Native AOT

Proven on a stripped ELF binary with no CoreCLR anywhere — `PublishAot` inside the csproj, gcc as the linker, a context over a type with all four interface kinds:

```
written : A4644C6973748201026353657481636F6E65634D6170A161610163536571820304
types   : List`1 HashSet`1 Dictionary`2 List`1
values  : [1,2] [one] a=1 [3,4]
ROUND TRIP OK
```

## Verification

**2029 library tests and 49 generator tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped; four TFMs build. Rebased onto `25d05a8`, and it adds **no build warnings** — checked deliberately, since that commit had just brought the solution to zero.

---

### DIFF
--- a/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
+++ b/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
@@ -393,6 +393,31 @@ public partial class Context : CborSerializerContext { }
         }
 
         /// <summary>
+        /// The collection interfaces the reflection path resolves through
+        /// <c>InterfaceCollectionConverter</c>, and <c>IDictionary&lt;K,V&gt;</c> through
+        /// <c>InterfaceDictionaryConverter</c>. Declaring one is ordinary — a model that says
+        /// <c>IList&lt;T&gt;</c> rather than <c>List&lt;T&gt;</c> is expressing a contract, not being
+        /// awkward — so refusing it made the generated path unusable for that model.
+        /// </summary>
+        [Theory]
+        [InlineData("System.Collections.Generic.IList<int>")]
+        [InlineData("System.Collections.Generic.ICollection<int>")]
+        [InlineData("System.Collections.Generic.IEnumerable<int>")]
+        [InlineData("System.Collections.Generic.IReadOnlyList<int>")]
+        [InlineData("System.Collections.Generic.IReadOnlyCollection<int>")]
+        [InlineData("System.Collections.Generic.ISet<int>")]
+        [InlineData("System.Collections.Generic.IDictionary<string, int>")]
+        public void ACollectionInterfaceMemberIsNotReported(string memberType)
+        {
+            AssertClean($@"
+public class Holder {{ public {memberType} Value {{ get; set; }} }}
+
+[CborSerializable(typeof(Holder))]
+public partial class Context : CborSerializerContext {{ }}
+");
+        }
+
+        /// <summary>
         /// An abstract base is never instantiated — its subtypes are — so it must not be reported.
         /// </summary>
         [Fact]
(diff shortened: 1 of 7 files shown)


Rafael-SOWNet:
Worth flagging before either of these merges: **#238 and #239 conflict with each other, in `GeneratedCorpusTests.Sample` only.** Both add a `Sample` entry at the same spot, and both entries are independent, so whichever merges second wants a rebase keeping both sides rather than the merge button.

That file is the standing magnet — every PR that declares a new context touches it — so the conflict is mechanical rather than a disagreement about the code. Nothing else in the two overlaps: #238 is the collection interfaces in `TypeCollector`/`Emitter`/`CddlTypeReference`, #239 is the two member features in `TypeCollector`/`Emitter`, and they do not share a line.

I will rebase whichever is second as soon as the first lands.

---

PR #239, opened 2026-08-16
Emit [CborRequired] and ShouldSerialize, which the mappings already supported

Rafael-SOWNet:
Two member features that were `CBOR1007` — "source generation does not reproduce this" — and are calls `DelegateMemberMapping` has always offered:

```csharp
.SetRequired(RequirementPolicy.Always)
.SetShouldSerializeMethod(o => ((Holder)o).ShouldSerializeName())
```

So the refusal was about the emitter not making a call, rather than a shape that cannot be reproduced. Both come off the `UnsupportedMemberFeatures` list.

## Two details worth the review time

**The requirement policy is resolved on the enum's own symbol, not mapped from its number.** That is not fastidiousness — `RequirementPolicy` is `Never = 0, Always = 1, AllowNull = 2, DisallowNull = 3`, which is not the order a reader assumes, and the hand-written table I wrote first got *every* value wrong. `RenderEnumConstant` walks the enum's fields for the matching constant and emits the fully qualified member, so a reordering cannot silently turn one policy into another here.

**`ShouldSerialize` repeats `DefaultObjectMappingConvention.ProcessShouldSerializeMethod`'s test rather than approximating it** — public, no parameters, returning `bool`, named for the member. A method the reflection path ignores has to be ignored here too: picking up a looser set would make the two paths disagree about whether a member is *written at all*, which is a silent divergence rather than a loud one. `AShouldSerializeMethodTheReflectionPathIgnoresIsNotUsed` pins it with a method that takes a parameter.

## Why the corpus cannot cover these

`GeneratedCorpusTests` compares one sample through both paths, and a sample that satisfies the requirement and passes the predicate encodes **identically whether or not either feature is honoured**. Both types are enrolled anyway — that is still worth having for the ordinary case — but the tests that bind use the values the features exist for:

- a document that *omits* the required member, asserted to throw on the generated path **and** on the reflection path, so it is a match rather than an opinion of its own;
- a value the predicate *excludes*, asserted to write a map of one and to equal the reflection path's bytes.

## Verification

**2032 library tests and 43 generator tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped; four TFMs build; **no new build warnings**, checked because `25d05a8` had just brought the solution to zero.

Binding measured by reverting **only the emission** — leaving the diagnostics removed so the test project still compiles, which the usual revert would not. That fails exactly the two behavioural tests and leaves the two positive controls passing, which is the split it should have.

Stacked on nothing: this is independent of #238, and the two do not touch the same lines.

---

### DIFF
--- a/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
+++ b/src/Dahomey.Cbor.Generator.Tests/DiagnosticTests.cs
@@ -93,7 +93,6 @@ public partial class Context : CborSerializerContext {{ }}
         }
 
         [Theory]
-        [InlineData("[CborRequired]")]
         [InlineData("[CborIgnoreIfDefault]")]
         [InlineData("[DefaultValue(3)]")]
         [InlineData("[CborLengthMode(LengthMode = LengthMode.IndefiniteLength)]")]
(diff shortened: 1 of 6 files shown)


Rafael-SOWNet:
Worth flagging before either of these merges: **#238 and #239 conflict with each other, in `GeneratedCorpusTests.Sample` only.** Both add a `Sample` entry at the same spot, and both entries are independent, so whichever merges second wants a rebase keeping both sides rather than the merge button.

That file is the standing magnet — every PR that declares a new context touches it — so the conflict is mechanical rather than a disagreement about the code. Nothing else in the two overlaps: #238 is the collection interfaces in `TypeCollector`/`Emitter`/`CddlTypeReference`, #239 is the two member features in `TypeCollector`/`Emitter`, and they do not share a line.

I will rebase whichever is second as soon as the first lands.

---

PR #245, opened 2026-08-16
Write a Half as a half-float, not as its own fields, closes #240

### LINKED ISSUE #240: A System.Half member is written as its internal fields rather than as a half-float
A member typed `System.Half` is serialized as an object — its own internal fields — rather than as a number. Measured on `25d05a8`, plain `new CborOptions()`, reflection path:

```csharp
public class Holder { public System.Half V { get; set; } }

Helper.Write(new Holder { V = (Half)1.5f });
```

```
A1 6156 A5
  6E 4269617365644578706F6E656E74 0F        "BiasedExponent": 15
  68 4578706F6E656E74 00                    "Exponent": 0
  6B 5369676E69666963616E64 190600          "Significand": 1536
  73 547261696C696E675369676E69666963616E64 190200
  66 5F76616C7565 193E00                    "_value": 15872
```

No exception, and nothing says anything. The document is meaningless to any other decoder, and to this one — reading it back depends on `Half`'s private layout.

## Why

`PrimitiveConverterProvider` has no case for `Half`, so it falls through to `ObjectConverterProvider` and `Half` is mapped like any other struct. Everything needed to encode it properly is already here: `CborWriter.WriteHalf`, `CborPrimitive.HalfFloat = 25`, and `HalfHelpers` for the bit conversions. `Half` resolves on all four target frameworks — `CborWriter.WriteHalf(Half)` is compiled without an `#if` — so a converter would not need conditional compilation.

RFC 8949 §3.3 gives it the obvious encoding: major type 7, additional value 25, the IEEE 754 binary16 the writer already emits for a `float` that round-trips through it.

## The generated path is not affected, and says so

A `Half` member on a `[CborSerializable]` type is `CBOR1002`:

```
'System.Half' cannot be handled by CBOR source generation (System.Half has no concrete converter
in PrimitiveConverterProvider, so the reflection path builds one through MakeGenericType; register
a custom converter for it)
```

That refusal is the better behaviour of the two — it is loud, and it is right that the generator will not reproduce a shape the reflection path gets wrong. Adding the converter closes both at once: `Half` becomes a primitive to the generator and the diagnostic stops being reachable.

## What I have not decided for you

Whether a `Half` should *read* only from a half-float, or from any numeric item the way `float` and `double` do — the writer is unambiguous but the reader's tolerance is a policy call, and it is the half of this that a caller's documents depend on.

Happy to open the PR once you have said which.

Rafael-SOWNet:
Closes #240.

`PrimitiveConverterProvider` had no case for `System.Half`, so it fell through to `ObjectConverterProvider` and was mapped like any other struct:

```
before : A1 6156 A5 6E4269617365644578706F6E656E74 0F … 665F76616C7565 193E00
         {"V": {"BiasedExponent":15,"Exponent":0,"Significand":1536,
                "TrailingSignificand":512,"_value":15872}}      -- all five of Half's own members
after  : A1 6556616C7565 F93E00
         {"Value": 1.5}
```

No exception, and a document no other decoder can read. **It did not round-trip either**: all five of those members are get-only or `readonly`, so a read could set none of them and returned `default`. The old encoding was write-only, which is what makes this change safe — nothing that round-tripped before round-trips differently now, because nothing round-tripped.

Such a document now fails with `Invalid major type Map` rather than silently yielding zero. That is a behaviour change and `docs/numbers.md` carries it as a note.

## The open question answered itself

I asked in the issue whether a `Half` should read only from a half-float or from any numeric item the way `float` does. **The library had already decided**: `CborReader.ReadHalf()` exists and is tolerant in exactly the way `ReadSingle` is. The converter defers to it and to `CborWriter.WriteHalf` in both directions, so a `Half` member is precisely as tolerant of what a sender chose as a `float` member is, rather than acquiring a policy of its own. A theory covers an integer, a negative integer, binary32 and binary64, plus saturation to infinity past binary16's range.

Nothing new was needed — `WriteHalf`, `ReadHalf` and `CborPrimitive.HalfFloat = 25` were all already there, and `Half` resolves on all four target frameworks, so there is no conditional compilation.

## Two consequential changes

**The generator stops refusing `Half`.** That refusal was the *better* of the two behaviours while the converter was missing, which is exactly why the generator's "no concrete converter" list is kept in step with `PrimitiveConverterProvider` by hand rather than derived from it: the reflection path did not refuse a `Half`, it wrote the struct. `Half[]` was already reachable through the RFC 8746 typed-array path; the scalar was the gap.

**CDDL renders a scalar `Half` as `float16`**, which is exact rather than merely permitted — this converter writes binary16 and nothing wider, where the single/double row needs the prelude's looser `float` because those emit the shortest form that round-trips. `ScalarHalfHasNoCddlRepresentation` becomes `ScalarHalfRendersAsFloat16`; flagging it because a test changing sides is the one thing not visible in the diff.

## Verification

**2050 library tests and 44 generator tests green on net10.0** with `CDDL_REQUIRED=1`, 0 skipped; four TFMs build; no new warnings.

`HalfHolder` is in `GeneratedCorpusTests`, so the generated and reflection paths are held to the same bytes rather than only to a round trip.

## Holding the provider and the collector in step

This change edits `PrimitiveConverterProvider` **and** `TypeCollector.IsPrimitive`, which are a matched pair that cannot share a list — the generator is an analyzer assembly and must not reference the runtime library. `PrimitiveConverterProviderParityTests` is new and exists for that: only a test can catch a one-sided edit.

The dangerous direction is the silent one. A type the collector omits while the provider resolves it is not *refused* — it is a struct, so it falls to `TypeKind.Object` and the generated context registers an `ObjectConverter` over members the reflection path never writes: a green build and the wrong bytes. That is why the assertion reads the emitted source rather than only the diagnostics.

Both guards this PR adds are proven to bind, by making the edit and running it:

* weakening `float16` to `float` fails `AHalfIsBinary16RatherThanTheWholeFloatFamily`;
* dropping `Half` from the collector alone fails the parity test.

The `float16` assertion also moved to `CddlScalarMappingTests`, which reaches the reference `cddl` tool — it had first landed in the diagnostics file, which never does, so the row had neither a parse check nor an instance check.

`README.md` and `docs/numbers.md` document `Half` now.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ Where the discriminator is written, when it is written, and how to change its na
 | [Object mapping](docs/object-mapping.md) | `ObjectMappingRegistry`, `AutoMap`, adjusting a single member |
 | [Polymorphism](docs/polymorphism.md) | Discriminators, policies and conventions |
 | [Typed arrays](docs/typed-arrays.md) | RFC 8746, `TypedArrayMode`, and what reads what |
-| [Numbers](docs/numbers.md) | `BigInteger`, `decimal`, `CborDecimalFraction` and `CborBigFloat` |
+| [Numbers](docs/numbers.md) | `BigInteger`, `decimal`, `CborDecimalFraction`, `CborBigFloat` and `Half` |
 | [Strings](docs/strings.md) | Indefinite-length strings, and why tag 25 is not supported |
 | [Tuples](docs/tuples.md) | `ValueTuple` at any arity, and the Native AOT requirement |
 | [Deterministic encoding](docs/deterministic-encoding.md) | RFC 8949 §4.2, key ordering, and hashing |
--- a/docs/numbers.md
+++ b/docs/numbers.md
@@ -2,8 +2,8 @@
 
 [← back to the README](../README.md)
 
-Three encodings beyond the basic integers and floats: bignums (tags 2 and 3), decimal fractions
-(tag 4) and bigfloats (tag 5).
+Three encodings beyond the basic integers and floats — bignums (tags 2 and 3), decimal fractions
+(tag 4) and bigfloats (tag 5) — and one at the other end of the range, half-precision floats.
 
 ## Bignums (RFC 8949 §3.4.3)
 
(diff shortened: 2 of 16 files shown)


Rafael-SOWNet:
Pushed `4b9f23d` after a review pass, and **corrected a count in the description**: `Half` has *five* unsettable members, not the four I listed — I had omitted `TrailingSignificand`.

That correction matters beyond arithmetic. All five are get-only or `readonly`, so the old encoding could not be read back at all: it was write-only, not merely unreadable by other decoders. Which is the real argument that this change is safe — nothing that round-tripped before round-trips differently now, because nothing round-tripped. Measured, not recalled.

Also in the push: `PrimitiveConverterProviderParityTests`, because this change edits the provider *and* the generator's `IsPrimitive` and those cannot share a list; the `float16` assertion moved to the CDDL file that actually reaches the gem; and `README.md`/`docs/numbers.md` now document `Half`, including the encoding change as a note.

Both new guards are proven to bind by making the one-sided edit each exists to catch. 2050 + 44 green, four TFMs, no new warnings.

